Issue #3, opened 2021-04-14
Initial scaffolding for Orb-Api and Orb-Agent


CheRuisiBesares:
The goal here is to establish a working skeleton for the `orb-api` and `orb-agent` and generate something concrete to iterate on.

PR #4, opened 2021-04-14
Initial Framework Proposal

CheRuisiBesares:
Proposal for #3 

This uses cobra and viper to manage cmd configs for both `orb-agent` and `orb-api`.

`Orb-agent` Uses the gin framework to setup the basic http api.

ToDos and Questions:

- [ ] Debug logger for gin framework doesn't log in a structured way. We can disable or fix this.
- [ ] Typed Datastores is maybe something for later at this point. Left Consul in a the default
- [ ] Communication with consul is still not hooked up
- [ ] Access Log json format is not well structured


CheRuisiBesares:
Going to put some inline discussion points for us to talk about

cmd/orb-agent/main.go
```diff
@@ -0,0 +1,97 @@
...
+		Short: "orb-agent connects to orb control plane",
+		Long:  "orb-agent connects to orb control plane",
+		Run:   Run,
+	}
+)
+
+func Run(cmd *cobra.Command, args []string) {
+	var config agent.Config
+	viper.Unmarshal(&config)
+	fmt.Printf("%+v\n", config)
```
CheRuisiBesares, line 43: We can remove this later but it just dumps the config out on start for now.

cmd/orb-agent/main.go
```diff
@@ -0,0 +1,97 @@
...
+
+func mergeOrError(path string) {
+	v := viper.New()
+	v.SetConfigFile(path)
+	cobra.CheckErr(v.ReadInConfig())
+
+	var fZero float64
+
+	// check that version of config files are all matched up
+	if versionNumber1 := viper.GetFloat64("version"); versionNumber1 != fZero {
```
CheRuisiBesares, line 66: `orb-agent` expects to see multiple config files on start so something like

`orb-agent -c conf1.yml conf2.yml` `mergeOrError` basically tries to merge the keys overwriting with the later config files on conflict. This checks that the version string of all the files presented match or it errors. If we like this functionality we can comment this and maybe add a quick test.

cmd/orb-api/config.example.yml
```diff
@@ -0,0 +1,26 @@
+listener:
+  bind_addr: "0.0.0.0:8080"
+  tls: 
+    auto:
```
CheRuisiBesares, line 4: This uses autotls when specified and will secure a cert using autocert at the supplied domain name.

cmd/orb-api/config.example.yml
```diff
@@ -0,0 +1,26 @@
+listener:
+  bind_addr: "0.0.0.0:8080"
+  tls:
```
CheRuisiBesares, line 3: If this section is empty the api will start in plain text. If auto is specified it will try to use autocert which requires a public ip. If you specify a cert and key path it will prefer those over anything else.

server/api/v1/example.go
```diff
@@ -0,0 +1,53 @@
+package v1
+
+import (
+	"github.com/gin-gonic/gin"
+)
+
+// we could make one of these per service or one for all of v1 depending on coupling points
```
CheRuisiBesares, line 7: Im not sure how exactly we want to structure this but i think we will know more when we get rolling more.

server/config.go
```diff
@@ -0,0 +1,59 @@
...
+	Format string `mapstructure:"format"`
+	Path   string `mapstructure:"path"`
+}
+
+type Loggers struct {
+	ErrorLog  ErrorLog  `mapstructure:"error_log"`
+	AccessLog AccessLog `mapstructure:"access_log"`
+}
+
+type Config struct {
```
CheRuisiBesares, line 55: I added this config to give the config file structure that is enforced. So we aren't really using viper to its fullest here because it can query the configs at runtime if we want. Im not sure which way we prefer to go here. Either way my plan was not to pass this config down into the api handlers and just let the server router assemble the needed components later.

server/router.go
```diff
@@ -0,0 +1,10 @@
+package server
+
+import (
+	"github.com/ns1labs/orb/server/api/v1"
+)
+
+func (s *Server) RegisterRoutes() {
+	v1RouteGroup := s.engine.Group("/v1")
```
CheRuisiBesares, line 8: We can add middleware to the v1 group when we have a little more clarity on what that looks like

server/server.go
```diff
@@ -0,0 +1,107 @@
...
+	s.closer = func() {
+		accessLogFile.Close()
+		errorLogFile.Close()
+	}
+
+	return s, nil
+
+}
+
+func (s *Server) Serve() error {
```
CheRuisiBesares, line 82: Use tls certs if supplied. Use autotls if supplied. Run in plain text on fall through.

server/server.go
```diff
@@ -0,0 +1,107 @@
...
+}
+
+type Server struct {
+	config Config
+	logger *log.Logger
+	engine *gin.Engine
+	closer func()
+}
+
+func (s *Server) Close() {
```
CheRuisiBesares, line 28: Im guessing there is a better way to do this but I wanted to make sure that the sever setup was distinct from the Serve Function while not leaving anything open? I can look at this closer for us but there is probably a smarter way.

CheRuisiBesares: Probably we should just collapse this into the Serve function and maybe leave New pretty basic

weyrick (approved):
looks like a great start, just the license thing and sketching in the test we discussed

PR #61, opened 2021-07-27
Resolves issues #50 #52 #53 #55 - Sink management

### LINKED ISSUE #50: Delete Sink

### LINKED ISSUE #52: Read for Sink resource

### LINKED ISSUE #53: Edit for Sink resource

### LINKED ISSUE #55: Create for Sink resource

dscabral:
Resolves #50, resolves #52, resolves #55, resolves #53


cmalerba (approved):
LGTM

sinks/service.go
```diff
@@ -11,31 +11,39 @@ package sinks
...
 )
 
-var (
-	ErrCreateSink = errors.New("failed to create sink")
-)
-
 type Service interface {
-	// CreateSink creates new data sink
-	CreateSink(ctx context.Context, token string, s Sink) (Sink, error)
+	SinkService
```
weyrick, line 23: I think this wrapper probably isn't necessary, and we should just use SinkService instead of Service unless you can think of a use case?

sinks/sinks_service.go
```diff
@@ -0,0 +1,122 @@
...
+	if backend.HaveBackend(sink.Backend) {
+		err = backend.GetBackend(sink.Backend).Connect(sink.Config)
+		if err != nil {
+			sink.Status = "not connected"
+			sink.Error = fmt.Sprint(err)
+		} else {
+			sink.Status = "connected"
+			sink.Error = ""
+		}
+	}
```
weyrick, line 43: it needs to be an error if the back end does not exist. this is where we will also use the ValidateOnly flag - if that is true, it should return here before saving

sinks/sinks_service.go
```diff
@@ -0,0 +1,122 @@
...
+	"github.com/ns1labs/orb/sinks/backend"
+)
+
+var (
+	ErrCreateSink                 = errors.New("failed to create Sink")
+	ErrConflictSink               = errors.New("entity already exists")
+	ErrUnsupportedContentTypeSink = errors.New("unsupported content type")
+)
+
+func (svc sinkService) CreateSink(ctx context.Context, token string, sink Sink) (Sink, error) {
```
weyrick, line 25: this probably needs a `validateOnly bool` parameter to be passed in from the API endpoint so we know if we are saving or just validating

sinks/api/endpoint.go
```diff
@@ -28,21 +29,196 @@ func addEndpoint(svc sinks.Service) endpoint.Endpoint {
...
-			Config: req.Config,
+			Name:        nID,
+			Backend:     req.Backend,
+			Config:      req.Config,
+			Description: req.Description,
+			Tags:        req.Tags,
+			Status:      req.Status,
+			Error:       req.Error,
 		}
 		saved, err := svc.CreateSink(ctx, req.token, sink)
```
weyrick, line 40: this is where we should pass in `req.ValidateOnly`

sinks/sinks_service.go
```diff
@@ -0,0 +1,122 @@
...
+
+func (svc sinkService) UpdateSink(ctx context.Context, token string, sink Sink) error {
+	skOwnerID, err := svc.identify(token)
+	if err != nil {
+		return err
+	}
+
+	sink.MFOwnerID = skOwnerID
+
+	if backend.HaveBackend(sink.Backend) {
```
weyrick, line 61: changing the backend type is not allowed, so this should not be needed. we should make sure that they can only update tags, description, and the other fields which are not read only

sinks/sinks_service.go
```diff
@@ -0,0 +1,122 @@
...
+		} else {
+			sink.Status = "connected"
+			sink.Error = ""
+		}
+	}
+	return svc.sinkRepo.Update(ctx, sink)
+}
+
+func (svc sinkService) ListBackends(ctx context.Context, token string) ([]string, error) {
+	_, err := svc.auth.Identify(ctx, &mainflux.Token{Value: token})
```
weyrick, line 75: why not use `svc.identify()` for these methods?

sinks/api/endpoint.go
```diff
@@ -28,21 +29,196 @@ func addEndpoint(svc sinks.Service) endpoint.Endpoint {
...
 
 		sink := sinks.Sink{
-			Name:   nID,
-			Config: req.Config,
+			Name:        nID,
+			Backend:     req.Backend,
+			Config:      req.Config,
+			Description: req.Description,
+			Tags:        req.Tags,
+			Status:      req.Status,
```
weyrick, line 37: Status and Error are read only so should not appear here

sinks/api/endpoint.go
```diff
@@ -28,21 +29,196 @@ func addEndpoint(svc sinks.Service) endpoint.Endpoint {
...
+		req := request.(updateSinkReq)
+		if err := req.validate(); err != nil {
+			return nil, err
+		}
+		nameID, _ := types.NewIdentifier(req.Name)
+		sink := sinks.Sink{
+			Name:        nameID,
+			ID:          req.id,
+			Tags:        req.Tags,
+			Status:      req.Status,
```
weyrick, line 72: Status, Backend, and Error are read only for Update so should not appear here

sinks/api/endpoint_test.go
```diff
@@ -0,0 +1,801 @@
...
+	nameID, _ = types.NewIdentifier("my-sink")
+	sink      = sinks.Sink{
+		Name:        nameID,
+		Description: "An example prometheus sink",
+		Backend:     "prometheus",
+		Config:      map[string]interface{}{"remote_host": "data", "username": "dbuser"},
+		Tags:        map[string]string{"cloud": "aws"},
+	}
+	invalidName  = strings.Repeat("m", maxNameSize+1)
+	notFoundRes  = toJSON(errorRes{things.ErrNotFound.Error()})
```
weyrick, line 53: this should be `sinks.ErrNotFound` instead of `things.ErrNotFound`, same for line 54

sinks/api/endpoint_test.go
```diff
@@ -0,0 +1,801 @@
...
+		req.Header.Set("Authorization", tr.token)
+	}
+	if tr.contentType != "" {
+		req.Header.Set("Content-Type", tr.contentType)
+	}
+	return tr.client.Do(req)
+}
+
+func newService(tokens map[string]string) sinks.Service {
+	auth := thmocks.NewAuthService(tokens)
```
weyrick, line 83: this brings in a dependency on mainflux things, we should probably fork this if we need it

sinks/api/endpoint_test.go
```diff
@@ -0,0 +1,801 @@
...
+			location:    "/sinks",
+		},
+		{
+			desc:        "add a valid without content type",
+			req:         validJson,
+			contentType: "",
+			auth:        token,
+			status:      http.StatusUnsupportedMediaType,
+			location:    "/sinks",
+		},
```
weyrick, line 158: let's add a test for ValidateOnly functionality

sinks/api/endpoint_test.go
```diff
@@ -0,0 +1,801 @@
...
+			auth:        token,
+			status:      http.StatusBadRequest,
+		},
+		"update sink with different owner": {
+			req:         invalidJson,
+			id:          sk.ID,
+			contentType: contentType,
+			auth:        token,
+			status:      http.StatusBadRequest,
+		},
```
weyrick, line 276: let's add a test to make sure you cannot update the read only fields: backend, status, error

sinks/api/endpoint_test.go
```diff
@@ -0,0 +1,801 @@
...
+	}
+}
+
+func TestViewBackend(t *testing.T) {
+	service := newService(map[string]string{token: email})
+	server := newServer(service)
+	defer server.Close()
+
+	bes, err := service.ListBackends(context.Background(), token)
+	//bes, err := service.ListBackends(context.Background(), token)
```
weyrick, line 508: remove

sinks/api/requests.go
```diff
@@ -35,3 +51,111 @@ func (req addReq) validate() error {
...
 	return nil
 }
+
+type updateSinkReq struct {
+	Name         string         `json:"name,omitempty"`
+	Backend      string         `json:"backend,omitempty"`
+	Config       types.Metadata `json:"config,omitempty"`
+	Description  string         `json:"description,omitempty"`
+	Tags         types.Tags     `json:"tags,omitempty"`
+	Status       string         `json:"status,omitempty"`
```
weyrick, line 61: remove the read only fields

sinks/api/transport.go
```diff
@@ -7,30 +7,81 @@ package api
...
+		types.EncodeResponse,
+		opts...,
+	))
+	r.Get("/features/sinks", kithttp.NewServer(
+		kitot.TraceServer(tracer, "list_backends")(listBackendsEndpoint(svc)),
+		decodeListBackends,
+		types.EncodeResponse,
+		opts...,
+	))
+	r.Get("/features/sinks/:id", kithttp.NewServer(
```
weyrick, line 67: in the original openapi spec this endpoint did not exist, it listed all information in the `/features/sinks` endpoint. I am fine to do it this way but we would need to update the openapi spec to match

weyrick (changes requested):
Great first pass! Please take a look and lmk if there are any questions on my comments

sinks/service.go
```diff
@@ -35,12 +31,10 @@ type PageMetadata struct {
 	Tags     types.Tags     `json:"tags,omitempty"`
 }
 
-var _ Service = (*sinkService)(nil)
```
weyrick, line 27: we should leave this check here but change it to SinkService instead of Service

sinks/sinks_service.go
```diff
@@ -56,31 +49,24 @@ func (svc sinkService) UpdateSink(ctx context.Context, token string, sink Sink)
...
+
+	sink.MFOwnerID = skOwnerID
 	return svc.sinkRepo.Update(ctx, sink)
 }
 
 func (svc sinkService) ListBackends(ctx context.Context, token string) ([]string, error) {
-	_, err := svc.auth.Identify(ctx, &mainflux.Token{Value: token})
+	_, err := svc.identify(token)
 	if err != nil {
 		return []string{}, errors.Wrap(errors.ErrUnauthorizedAccess, err)
```
weyrick, line 63: these should just return `err`, it is already wrapped by `identify`

sinks/sinks_service.go
```diff
@@ -104,19 +90,35 @@ func (svc sinkService) ViewSink(ctx context.Context, token string, key string) (
...
+	res, err := svc.identify(token)
 	if err != nil {
 		return errors.Wrap(errors.ErrUnauthorizedAccess, err)
 	}
 
-	return svc.sinkRepo.Remove(ctx, res.GetId(), id)
+	return svc.sinkRepo.Remove(ctx, res, id)
+}
+
+func validateBackend(sink Sink) error {
```
weyrick, line 110: this parameter has to be passed by pointer since the goal is to modify the object. this was caught by LGTM static analysis: https://lgtm.com/projects/g/ns1labs/orb/rev/pr-70e8d1c4f6ce1cf6c73de8a1743874fd88ecaa22

weyrick (changes requested):
good stuff, left a few more comments

weyrick (approved):
let's do it 🔥

PR #87, opened 2021-08-05
feat(angular-prod-env): add api versioning to endpoint urls

### LINKED ISSUE #86: Configure Angular environment for --prod builds
When dev build, lookup urls normally since likely running with mock services or proxying docker services while developing with ng -serve.

--prod builds, such as docker image build, are not looking up the endpoints properly, api versioning is missing from endpoint urls

gpazuch:
Set up different endpoint urls based on --prod or default builds,
prepending orb defined endpoints url with *<api_url_prefix><version_number>*

eg.: `sinksUrl: '/sinks'` -> `sinksUrl: '/api/v1/sinks'`

closes #86

### DIFF
--- a/ui/src/environments/environment.defaults.ts
+++ b/ui/src/environments/environment.defaults.ts
@@ -1,30 +1,40 @@
-export const environment = {
-  usersUrl: '/users',
-  groupsUrl: '/groups',
-  membersUrl: '/members',
-  usersVersionUrl: '/version',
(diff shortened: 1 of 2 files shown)


gpazuch:
thx @dscabral for pointing out that dockerized version of UI didn't have the correct routes to endpoints. much appreciated

daitan-rovaris (approved):
lgtm!

PR #95, opened 2021-08-09
Resolves #83 Sink validation endpoint

### LINKED ISSUE #83: Sink validation endpoint

dscabral:
Resolves issue #83.


sinks/api/requests.go
```diff
@@ -154,3 +154,30 @@ func (req deleteSinkReq) validate() error {
...
 	return nil
 }
+
+type validateReq struct {
+	Name         string         `json:"name,omitempty"`
+	Backend      string         `json:"backend,omitempty"`
+	Config       types.Metadata `json:"config,omitempty"`
+	Description  string         `json:"description,omitempty"`
+	Tags         types.Tags     `json:"tags,omitempty"`
+	ValidateOnly bool           `json:"validate_only"`
```
weyrick, line 164: We should remove this now because we have the separate validate endpoint instead. Also remove from updateReq and addReq

sinks/api/openapi.yaml
```diff
@@ -139,6 +139,33 @@ paths:
           description: A non-existent entity request.
         '500':
           $ref: "#/components/responses/ServiceErrorRes"
+  /sinks/validate:
+    parameters:
+      - $ref: "#/components/parameters/Authorization"
+    post:
+      summary: "validate a new sink"
```
weyrick, line 146: "validate a sink configuration without saving it"

sinks/sinks.go
```diff
@@ -91,4 +93,6 @@ type SinkRepository interface {
 	RetrieveById(ctx context.Context, key string) (Sink, error)
 	// Remove a existing Sink by id
 	Remove(ctx context.Context, owner string, key string) error
+	// RetrieveToValidate validates a sink
+	RetrieveToValidate(ctx context.Context, name, owner string) error
```
weyrick, line 97: why is this needed?

sinks/postgres/sinks.go
```diff
@@ -196,6 +196,29 @@ func (s sinksRepository) Remove(ctx context.Context, owner, id string) error {
 	return nil
 }
 
+func (s sinksRepository) RetrieveToValidate(ctx context.Context, name, owner string) error {
```
weyrick, line 199: why is this needed?

weyrick (changes requested):
I don't think a repository method is needed here, and we still need to resolve this branch so that no `.ts` files have any changes

sinks/sinks.go
dscabral: We also check if the name already exists on database. So, before we save it, we can check if the name is available, and not duplicated.

lterranova:
I'm closing this PR so we can fix the problem with multiple files coming from the frontend.
I'm going to create a new one with all the changes requested by @weyrick.

PR #104, opened 2021-08-12
fix(orb-ui env.prod-angular): wrong login endpoint. not versioned yet.

gpazuch:
* fix loginUrl '/tokens/api/v1' -> '/tokens'

Revert versioning of tokens endpoint


daitan-rovaris (approved):
lgtm

PR #105, opened 2021-08-12
feat(orb-sink orb-ui sink-dashboard): improve css and icons of sink d…

gpazuch:
* improve most aspects close do design spec
* color, fonts and sizes of texts, borders and background close to or equal spec.
* improve icons - not final icons
* add breadcrumb do sink dashboard
* fix table width*height
* table column sizes not to spec.

Improves Sink Dashboard UI a lot closer to spec.
Sink table still not up to spec. Table component to be replaced.


daitan-rovaris (approved):
lgtm

PR #120, opened 2021-08-16
feat(sink): create new config fields to available backends

### LINKED ISSUE #121: Sink Management: Create new fields to available backends

dscabral:
Resolves #121 
- Create new fields to backend configuration 
- Remove content type validation to GET requests

### DIFF
--- a/sinks/backend/prometheus/prometheus.go
+++ b/sinks/backend/prometheus/prometheus.go
@@ -22,17 +22,9 @@ type prometheusBackend struct {
 }
 
 type SinkFeature struct {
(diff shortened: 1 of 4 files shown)


sinks/backend/prometheus/prometheus.go
```diff
@@ -65,3 +70,31 @@ func Register() bool {
 
 	return true
 }
+
+func createFeatureConfig() []configFeature {
```
weyrick, line 74: looks good, this method should probably be part of the Backend interface, as all back ends will implement it right?

weyrick (approved):

PR #170, opened 2021-08-25
refactor(orb-ui sinks#table): remove hover event 

gpazuch:
remove hover event of c[RUD] button group of each row

* remove `*ngIf="row['isHover']` from button group of each sink row.

Table to be replaced before the weekend anyways, just to make testing easier now


daitan-rovaris (approved):
lgtm

PR #228, opened 2021-09-07
feat(orb-ui orb-docker): Dockerfile update npm to yarn

gpazuch:
Update [Dockerfile](./ui/docker/Dockerfile) to use yarn instead of npm.
Yarn resolves peer deps and package conflicts with more ease.
Also quicker than update the giant package.json dep list RN.

Update Orb UI dockerfile commands

### DIFF
--- a/ui/docker/Dockerfile
+++ b/ui/docker/Dockerfile
@@ -2,9 +2,12 @@
 FROM node:14.15.4 as node
 WORKDIR /app
 COPY package.json /app/
-RUN npm install
+# I'm not sure if yarn is available from this node it imported
+# but it could be installed with npm install -g yarn
+# RUN npm install -g yarn
+RUN yarn install
 COPY ./ /app/
-RUN npm run build:prod
+RUN yarn build:prod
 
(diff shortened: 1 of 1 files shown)


daitan-rovaris (changes requested):
Hey @gpazuch please add the RUN npm install -g yarn command, to be sure it will be there

ui/docker/Dockerfile
```diff
@@ -2,9 +2,12 @@
 FROM node:14.15.4 as node
 WORKDIR /app
 COPY package.json /app/
-RUN npm install
+# I'm not sure if yarn is available from this node it imported
```
weyrick, line 5: yes it's available, let's remove these comments and then we can land it

daitan-rovaris: > yes it's available, let's remove these comments and then we can land it

:+1:

gpazuch: You're right, it does indeed come pre-installed.
@daitan-rovaris I've just removed the --force;
After running both alternatives and checking yarn versions available, it seemed to be overkill.

daitan-rovaris (approved):
lgtm!

PR #229, opened 2021-09-07
wip - feat(orb-ui): Agent Add Page

gpazuch:
Add Agent Add Page

Location field is extracted and set to orb_tags['location'] upon creation and edition

Inherited tags are still not being displayed


daitan-rovaris (approved):
lgtm

PR #230, opened 2021-09-08
Feature/orb edit-delete policies

### LINKED ISSUE #204: Policy Update
Update: if policy is used in any datasets, send policy update RPC over group channel ID for each dataset

### LINKED ISSUE #205: Policy Delete

dscabral:
Resolves #205, resolves #204


policies/api/http/transport.go
```diff
@@ -42,21 +43,31 @@ func MakeHandler(tracer opentracing.Tracer, svcName string, svc policies.Service
 	}
 	r := bone.New()
 
-	r.Post("/policies/agent", kithttp.NewServer(
+	r.Post("/policies/agents", kithttp.NewServer(
```
weyrick, line 46: this should stay `agent` singular. sorry I think you must have looked at the architecture diagram, which erroneously does say `agents`, we need to fix that. our openapi though is singular, and the text on the architecture page is singular.

policies/comms.go
```diff
@@ -0,0 +1,33 @@
+// Copyright (c) Mainflux
```
weyrick, line 1: We discussed this should not be needed - policies we'll never communicate with agents directly. It looks like it may be unused and is old code?

policies/comms_rpc.go
```diff
@@ -0,0 +1,13 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
```
weyrick, line 1: Same, this should not be necessary/used

fleet/api/grpc/endpoint.go
```diff
@@ -12,11 +12,30 @@ func retrieveAgentEndpoint(svc fleet.Service) endpoint.Endpoint {
...
 		}
-		agentGroup, err := svc.ViewAgentGroupByIDInternal(ctx, req.OwnerID, req.AgentID)
+		agent, err := svc.ViewAgentByIDInternal(ctx, req.OwnerID, req.AgentID)
 		if err != nil {
 			return nil, err
 		}
 		res := agentRes{
+			id:      agent.MFThingID,
+			name:    agent.Name.String(),
+			channel: agent.MFChannelID,
```
weyrick, line 22: I think you set this up to get the channel, but we know we don't need that sense policies will not be using it, so do we need this still?

fleet/redis/consumer/streams.go
```diff
@@ -108,6 +123,68 @@ func (es eventStore) handleDatasetCreate(ctx context.Context, e createDatasetEve
...
+		return updatePolicyEvent{}, err
+	}
+
+	return val, nil
+}
+
+// the policy service is notifying that a policy has been updated
+// notify all agents in the AgentGroup specified in the dataset about the policy update
+func (es eventStore) handlePolicyUpdate(ctx context.Context, e updatePolicyEvent) error {
+	// todo check if it will be necessary create a new comms function to notify a update
```
weyrick, line 147: this is actually a good question - I think the way you have it might be okay though (using NotifyGroupNewAgentPolicy)

policies/policy_service.go
```diff
@@ -126,3 +112,92 @@ func (s policiesService) ViewPolicyByID(ctx context.Context, token string, polic
...
+	}
+	pol.Backend = plcy.Backend
+	pol.MFOwnerID = ownerID
+
+	err = validatePolicyBackend(&pol, format, policyData)
+	if err != nil {
+		return Policy{}, err
+	}
+
+	err = s.repo.UpdatePolicy(ctx, ownerID, pol)
```
weyrick, line 135: Before saving to the repo we want to increase the `pol.Version` by 1

policies/redis/producer/streams.go
```diff
@@ -28,6 +36,94 @@ type eventStore struct {
...
+	err = validatePolicyBackend(&pol, format, policyData)
+	if err != nil {
+		return policies.Policy{}, err
+	}
+
+	event := updatePolicyEvent{
+		id:       pol.ID,
+		ownerID:  pol.MFOwnerID,
+		groupIDs: strings.Join(groupsIDs, ","),
+		policy:   pol.Policy,
```
weyrick, line 111: It seems like we aren't using this right now because the fleet service is making an RPC call back to policies to retrieve it during an edit. If we are going to do that for now, then let's remove this policy from redis because they could be large and waste CPU/memory.

weyrick (changes requested):
Ok looks pretty good, let's just clean up the unused "policy comms" code, and if we don't need these fleet gRPC functions for retrieving agent and agent group, get rid of them for now too (unless you know that we will need them later)

fleet/redis/consumer/streams.go
dscabral: Nice

policies/policy_service.go
dscabral: Done

policies/redis/producer/streams.go
dscabral: Ok, done.

weyrick (approved):

PR #233, opened 2021-09-08
fix(orb-ui orb-sinks): Fix #140 - unable to edit sink.

### LINKED ISSUE #140: Sink Management: Unable to edit existing sink
Edit option doesn't allow user to edit an existing sink

Steps to reproduce:

- open ORB on Sink Management page
- Press edit option for an existing sink
- Try to edit some information except name label

Expected Result:

- You should be able to edit sink and new information should overwrite the oldest.

Current Result:

- User isn't allowed to edit existing sink keeping the current name label
- If label name is changed, a new sink is being created 

Check the issue by clicking here: https://drive.google.com/file/d/1UBWtbMRGXe1plA_HuXBrG5RnPDBXCQHv/view?usp=sharing

gpazuch:
Fixes a few issues that have been introduced with the new routing in place.

*Fix routing from and to edit sink page.
*Fix agent fetch and form control patchData

Closes #140


daitan-rovaris (approved):
cleaner now! lgtm

Issue #240, opened 2021-09-09
Agent: Enabled to create agent without tag


manrodrigues:

manrodrigues:
![Orb (5)](https://user-images.githubusercontent.com/78241475/132618220-5b7d84fa-f0fd-415d-87c1-35bf681d9341.jpg)

manrodrigues:
Note that (location, location) referred to text inserted on location

iatdaitan:
This issue was already fixed.

Issue #237, opened 2021-09-09
Agent: Location is being displayed as an orb tag


manrodrigues:
Agent tags is empty and location is being displayed as an orb tag

manrodrigues:
![Orb](https://user-images.githubusercontent.com/78241475/132616663-da34a224-05e9-4528-a25a-e3d06fd7ba8f.jpg)

iatdaitan:
Location has been removed from the MVP.

PR #247, opened 2021-09-09
feat(orb-ui): Agent, Agent Group & Sink Fixes

### LINKED ISSUE #242: Agent: delete agent message refers to agent label, but we have just agent name
Note that the same problem is happening for the deletions of the agent groups and sinks

gpazuch:
Fixes a few issues with current implementation

* Routing back to list from create/update should be fine for all pages
* Misplaced strings should be ok
* Fix *allow* create|update with 0 tags - all entities must have at least one tag defined.  - location tag for agent does not count.
* Fix Agent Update correct values and tags filled in
* Fix Agent Tags showing location tag

closes #242 #240 #230 #237


dscabral (approved):

daitan-rovaris (approved):
looking good

PR #252, opened 2021-09-10
add taps from pktvisor to agent_metadata capabilities

weyrick:
Query pktvisor for tap information and send as Capabilities during orb agent connection.
Also fix TLS verify default, and move Heartbeat frequency to correct module.


dscabral (approved):
I got it... more simple than i though.

PR #257, opened 2021-09-10
fix: minor ui fixes

iatdaitan:
- Fixed summary message of agents, agent groups, and sinks
- Fixed Add new Sink button position
- Standardized tag format on all available pages
- Fixed show of Agent Tags and Orb Tags


gpazuch (approved):
lgtm!

PR #263, opened 2021-09-14
Add description to dockerhub  repos using README

jjvitorino:
Automatically add description and short description in the dockerhub repos using the content of README.

Right now, it uses only the README on the root of the project for all the DH repositories, but it is possible  to add a README for each one of them using `readme-filepath: ./path/to/README.md` in the pipeline.

### DIFF
--- a/.github/workflows/dockerhub-description.yml
+++ b/.github/workflows/dockerhub-description.yml
@@ -1,0 +1,78 @@
+
+
+name: Update Docker Hub Description
+on:
+  push:
+    branches: [ develop, main ]
+    paths: [ README.md, .github/workflows/dockerhub-description.yml ]
+  pull_request:    
+    branches: [ develop, main ]
+    paths: [ README.md, .github/workflows/dockerhub-description.yml ]
+  
+
+  
+jobs:
+  # There is no support for yaml anchor on Github_actions
+  # https://github.com/actions/runner/issues/1182 
+  dockerHubDescription:
+    runs-on: ubuntu-latest
+    steps:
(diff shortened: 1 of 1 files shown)


.github/workflows/dockerhub-description.yml
```diff
@@ -0,0 +1,78 @@
...
+    steps:
+    - uses: actions/checkout@v2
+
+    - name: Docker Hub Description agent
+      uses: peter-evans/dockerhub-description@v2
+      # The default readme is in the root of the repository but any file 
+      # can be used, example: readme-filepath: ./path/to/README.md
+      with: 
+        username: ${{ secrets.DOCKERHUB_USERNAME }}
+        password: ${{ secrets.DOCKERHUB_PASSWORD }}
```
weyrick, line 28: can we change this to DOCKERHUB_TOKEN? that is already defined.

weyrick (changes requested):
Thanks for this. Question - it is currently failing the action in your PR, saying username is not specified - but DOCKERHUB_USERNAME is defined for this repo...?

.github/workflows/dockerhub-description.yml
weyrick: Sorry I meant the `secrets` name for `password` is `DOCKERHUB_TOKEN` ... so:
```
        password: ${{ secrets.DOCKERHUB_TOKEN }}
```
I would expect that to work, are you saying it wouldn't?

jjvitorino: I'll dig more on this.

PR #279, opened 2021-09-14
Feature/orb dataset validation endpoint

### LINKED ISSUE #265: Dataset Validation

mclcavalcante:
Resolves #265


dscabral (changes requested):

dscabral (approved):

policies/policy_service.go
```diff
@@ -218,4 +219,32 @@ func (s policiesService) ValidatePolicy(ctx context.Context, token string, p Pol
...
+	}
+
+	d.MFOwnerID = mfOwnerID
+
+	_, err = uuid.FromString(d.SinkID)
+	if err != nil {
+		fmt.Println("invalid sink id")
+		return Dataset{}, ErrMalformedEntity
+	}
+
```
weyrick, line 236: In addition to making sure that the ID is a valid UUID, we also want to make sure that the ID represents a valid Sink. This means it should be making a gRPC call to ensure it can retrieve a Sink with that ID for this owner. The same goes for policy and group ID.

weyrick (changes requested):
In general this looks good except that we need to move to accepting an array of Sink IDs instead of a single Sink ID (which affects most of this code in some way) and to verify the IDs in the dataset represent valet objects (sinks, policy, group)

weyrick (approved):
Nice!

PR #287, opened 2021-09-15
fix(orb-ui): formControl name input validators

gpazuch:
Add name input validators for `name` fields of `Agent`, `AgentGroup` and `Sink` Orb entities.


iatdaitan (approved):
lgtm.

PR #318, opened 2021-09-22
feat(orb-ui): #270 Policy Create Page

gpazuch:
Add Policy Creation Page

Add Mock for policy creation for pktvisor


gpazuch:
#!WIP!#
Do not merge, it was supposed to be a draft, work in progress.

ui/src/app/common/services/agents/agent.policies.service.ts
```diff
@@ -167,4 +169,233 @@ export class AgentPoliciesService {
         },
       );
   }
+
```
gpazuch, line 172: @dscabral these are the mocked data I'm using for the agent policies service here in FE before we integrate with BE

PR #374, opened 2021-10-08
fix(orb-ui): validate datasets

gpazuch:


gpazuch:
[dataset](https://github.com/ns1labs/orb/blob/ee1665ddd5a62ecec1d6c643eb91707f53cad24a/policies/api/http/requests.go#L128)

[dataset](https://github.com/ns1labs/orb/blob/ee1665ddd5a62ecec1d6c643eb91707f53cad24a/policies/api/http/responses.go#L141)

@mclcavalcante pointed out to me another error on datasets, I'm passing `sink_id` instead of `sink_ids`

I'll update Dataset interface on FE and [wiki](https://github.com/ns1labs/orb/wiki/Architecture:-Policies-and-Datasets)

PR #414, opened 2021-11-01
Feature/handle dataset remove rpc

mclcavalcante:


agent/comms.go
```diff
@@ -69,6 +69,12 @@ func (a *orbAgent) unsubscribeGroupChannel(channelID string) {
 	a.logger.Info("completed RPC unsubscription to group", zap.String("topic", rpcFromCoreTopic))
 }
 
+func (a *orbAgent) removeDatasetFromPolicy(datasetID string, policyID string) {
+	for _, be := range a.backends {
```
weyrick, line 73: This shouldn't be sending to all backends, but we can't fix this until we refactor policy/dataset management so leave for now.

weyrick (changes requested):

weyrick (approved):

PR #417, opened 2021-11-02
fix(orb-ui): Sink tags are no longer required

### LINKED ISSUE #359: Sink Tags Should be Optional

gpazuch:
Fix -> sink tags are no longer required.
Close #359


weyrick (commented):
Tags should only ever be required on Agent Groups. Does this just fix sinks? Can we fix the others too?

PR #514, opened 2021-11-11
fix(orb-ui): #467 agent tag input

### LINKED ISSUE #467: Agents: Unable to exclude all tags from an agent already created
When edit an existing agent, removal of all tags is not enable (although I can create an agent without any tag). 
If removing all tags after the agent is tagged really can't happen, the interface should inform the user about it.

(note that if an agent is created without tags, edition of this agent turn into an 400- Bad request error)

gpazuch:
Resolves #467


gpazuch:
I could swear I had done this already.

This PR has been sitting here for a while, and it is now obselete I think. It is closed.

PR #531, opened 2021-11-16
feature(fleet): add metrics to api

mclcavalcante:


cmd/fleet/main.go
```diff
@@ -215,13 +215,13 @@ func newFleetService(auth mainflux.AuthServiceClient, db *sqlx.DB, logger *zap.L
 			Subsystem: "api",
 			Name:      "request_count",
 			Help:      "Number of requests received.",
-		}, []string{"method"}),
+		}, []string{"method", "owner_id", "service", "id"}),
```
weyrick, line 218: So I think best practice is to not put "service" here since it will never change (that can be added by the scraper). Also, I believe we should use "thing_id" instead of "id" here. This way we can clearly specify during queries which ID we mean, and it results in lower cardinality for the label since other services also have "id"

fleet/api/http/metrics.go
```diff
@@ -18,71 +19,306 @@ type metricsMiddleware struct {
...
+	}(time.Now())
+
 	return m.svc.ViewOwnerByChannelIDInternal(ctx, channelID)
 }
 
 func (m metricsMiddleware) ViewAgentBackend(ctx context.Context, token string, name string) (interface{}, error) {
+	defer func(begin time.Time) {
+		labels := []string{
+			"method", "viewAgentBackend",
+			"owner_id", "",
```
weyrick, line 43: If we don't have a value, I believe we should leave it out of this list, I think we don't have to pass an empty string? Also, we _should_ have owner_id here because we have a token. Let's discuss how to pull it out.

fleet/api/http/metrics.go
```diff
@@ -18,71 +19,306 @@ type metricsMiddleware struct {
...
+	defer func(begin time.Time) {
+		labels := []string{
+			"method", "viewOwnerByChannelIDInternal",
+			"owner_id", agent.MFOwnerID,
+			"service", "fleet",
+			"id", agent.MFThingID,
+		}
+
+		m.counter.With(labels...).Add(1)
+		m.latency.With(labels...).Observe(time.Since(begin).Seconds())
```
weyrick, line 32: Should Seconds be Microseconds?

cmd/fleet/main.go
```diff
@@ -209,19 +209,20 @@ func newFleetService(auth mainflux.AuthServiceClient, db *sqlx.DB, logger *zap.L
...
 			Help:      "Number of requests received.",
-		}, []string{"method", "owner_id", "service", "id"}),
+		}, []string{"method", "owner_id", "thing_id"}),
 		kitprometheus.NewSummaryFrom(stdprometheus.SummaryOpts{
 			Namespace: "fleet",
 			Subsystem: "api",
 			Name:      "request_latency_microseconds",
 			Help:      "Total duration of requests in microseconds.",
-		}, []string{"method", "owner_id", "service", "id"}),
+		}, []string{"method", "owner_id", "thing_id"}),
```
weyrick, line 225: so we will need to add "group_id" here too, since some of these calls are agent group and not just thing. see below.

fleet/api/http/metrics.go
```diff
@@ -126,12 +133,11 @@ func (m metricsMiddleware) ViewAgentGroupByIDInternal(ctx context.Context, group
 		labels := []string{
 			"method", "viewAgentGroupByIDInternal",
 			"owner_id", ownerID,
-			"service", "fleet",
-			"id", groupID,
+			"thing_id", groupID,
```
weyrick, line 136: here's the example of a groupID - we can't put this into thing_id, it needs its own.

weyrick (changes requested):

fleet/api/http/metrics.go
```diff
@@ -16,81 +19,329 @@ type metricsMiddleware struct {
...
 	svc     fleet.Service
+	auth    mainflux.AuthServiceClient
 }
 
-func (m metricsMiddleware) ViewOwnerByChannelIDInternal(ctx context.Context, channelID string) (fleet.Agent, error) {
+func (m metricsMiddleware) ViewOwnerByChannelIDInternal(ctx context.Context, channelID string) (agent fleet.Agent, _ error) {
+	defer func(begin time.Time) {
+		labels := []string{
+			"method", "viewOwnerByChannelIDInternal",
+			"owner_id", agent.MFOwnerID,
```
dscabral, line 29: Probably these fields are empty. You should change the order of the m.svc.ViewOwnerByChannelIDInternal, calling it before you span the metrics. You're using agent, but it's the returned object.

fleet/api/http/metrics.go
```diff
@@ -16,81 +19,329 @@ type metricsMiddleware struct {
...
+
 	return m.svc.ViewAgentByIDInternal(ctx, ownerID, thingID)
 }
 
-func (m metricsMiddleware) ViewAgentByID(ctx context.Context, token string, thingID string) (fleet.Agent, error) {
+func (m metricsMiddleware) ViewAgentByID(ctx context.Context, token string, thingID string) (a fleet.Agent, _ error) {
+	defer func(begin time.Time) {
+		labels := []string{
+			"method", "viewAgentByID",
+			"owner_id", a.MFOwnerID,
```
dscabral, line 103: Same here

fleet/api/http/metrics.go
```diff
@@ -16,81 +19,329 @@ type metricsMiddleware struct {
...
+
 	return m.svc.ViewAgentByID(ctx, token, thingID)
 }
 
-func (m metricsMiddleware) EditAgent(ctx context.Context, token string, agent fleet.Agent) (fleet.Agent, error) {
+func (m metricsMiddleware) EditAgent(ctx context.Context, token string, agent fleet.Agent) (a fleet.Agent, _ error) {
+	defer func(begin time.Time) {
+		labels := []string{
+			"method", "editAgent",
+			"owner_id", a.MFOwnerID,
```
dscabral, line 119: Same here

fleet/api/http/metrics.go
```diff
@@ -16,81 +19,329 @@ type metricsMiddleware struct {
...
+
 	return m.svc.ViewAgentGroupByIDInternal(ctx, groupID, ownerID)
 }
 
-func (m metricsMiddleware) ViewAgentGroupByID(ctx context.Context, token string, groupID string) (fleet.AgentGroup, error) {
+func (m metricsMiddleware) ViewAgentGroupByID(ctx context.Context, token string, groupID string) (group fleet.AgentGroup, _ error) {
+	defer func(begin time.Time) {
+		labels := []string{
+			"method", "viewAgentGroupByID",
+			"owner_id", group.MFOwnerID,
```
dscabral, line 151: Same here

fleet/api/http/metrics.go
```diff
@@ -16,81 +19,329 @@ type metricsMiddleware struct {
...
+
 	return m.svc.ValidateAgentGroup(ctx, token, s)
 }
 
-func (m metricsMiddleware) ValidateAgent(ctx context.Context, token string, a fleet.Agent) (fleet.Agent, error) {
+func (m metricsMiddleware) ValidateAgent(ctx context.Context, token string, a fleet.Agent) (agent fleet.Agent, _ error) {
+	defer func(begin time.Time) {
+		labels := []string{
+			"method", "validateAgent",
+			"owner_id", agent.MFOwnerID,
```
dscabral, line 294: Same here

fleet/api/http/metrics.go
```diff
@@ -16,81 +19,329 @@ type metricsMiddleware struct {
...
+
 	return m.svc.RemoveAgentGroup(ctx, token, groupID)
 }
 
-func (m metricsMiddleware) ValidateAgentGroup(ctx context.Context, token string, s fleet.AgentGroup) (fleet.AgentGroup, error) {
+func (m metricsMiddleware) ValidateAgentGroup(ctx context.Context, token string, s fleet.AgentGroup) (group fleet.AgentGroup, _ error) {
+	defer func(begin time.Time) {
+		labels := []string{
+			"method", "validateAgentGroup",
+			"owner_id", group.MFOwnerID,
```
dscabral, line 278: Same here

fleet/api/http/metrics.go
```diff
@@ -16,81 +19,329 @@ type metricsMiddleware struct {
...
+
 	return m.svc.CreateAgent(ctx, token, a)
 }
 
-func (m metricsMiddleware) CreateAgentGroup(ctx context.Context, token string, s fleet.AgentGroup) (fleet.AgentGroup, error) {
+func (m metricsMiddleware) CreateAgentGroup(ctx context.Context, token string, s fleet.AgentGroup) (group fleet.AgentGroup, _ error) {
+	defer func(begin time.Time) {
+		labels := []string{
+			"method", "createAgentGroup",
+			"owner_id", group.MFOwnerID,
```
dscabral, line 241: Same here

fleet/api/http/metrics.go
```diff
@@ -16,81 +19,329 @@ type metricsMiddleware struct {
...
+
 	return m.svc.ListAgents(ctx, token, pm)
 }
 
-func (m metricsMiddleware) CreateAgent(ctx context.Context, token string, a fleet.Agent) (fleet.Agent, error) {
+func (m metricsMiddleware) CreateAgent(ctx context.Context, token string, a fleet.Agent) (agent fleet.Agent, _ error) {
+	defer func(begin time.Time) {
+		labels := []string{
+			"method", "createAgent",
+			"owner_id", agent.MFOwnerID,
```
dscabral, line 225: Same here

fleet/api/http/metrics.go
```diff
@@ -16,81 +19,329 @@ type metricsMiddleware struct {
...
+	}(time.Now())
+
 	return m.svc.ListAgentGroups(ctx, token, pm)
 }
 
 func (m metricsMiddleware) EditAgentGroup(ctx context.Context, token string, ag fleet.AgentGroup) (fleet.AgentGroup, error) {
+	defer func(begin time.Time) {
+		labels := []string{
+			"method", "editAgentGroup",
+			"owner_id", ag.MFOwnerID,
```
dscabral, line 188: Same here

dscabral (changes requested):

fleet/api/http/metrics.go
```diff
@@ -17,53 +19,61 @@ type metricsMiddleware struct {
...
-		m.latency.With(labels...).Observe(time.Since(begin).Seconds())
+		m.latency.With(labels...).Observe(float64(time.Since(begin).Microseconds()))
 
 	}(time.Now())
 
 	return m.svc.ViewOwnerByChannelIDInternal(ctx, channelID)
 }
 
 func (m metricsMiddleware) ViewAgentBackend(ctx context.Context, token string, name string) (interface{}, error) {
+	ownerID, err := m.identify(token)
```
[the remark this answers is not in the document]
dscabral: Using this approach, yes.

fleet/api/http/metrics.go
weyrick: So we should verify this - because this is run `defer` I think there might be a value, since it should be running after the function is all finished

fleet/api/http/metrics.go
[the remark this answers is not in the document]
weyrick: Seems to me we could be using the result from the inner call (ViewAgentBackend in this case) - we just need to figure out how to access it. Maybe it can be stored somewhere

fleet/api/http/metrics.go
dscabral: You're right... just need to confirm. I think I face empty values on redis producer, but it was not running `defer`.

mclcavalcante: I had checked this earlier, here is an example, it's not the same function but the same approach and we're able get the value this way. See the value of index 3
image

dscabral: You're are completly right, I also tested here with same results. My bad... I did remove the other comments.

gpazuch:
Hi @dscabral , @mclcavalcante . 
I see some metrics are being gathered for api calls.
Can you point me to the service/endpoints to retrieve those metrics?

weyrick:
> Hi @dscabral , @mclcavalcante . I see some metrics are being gathered for api calls. Can you point me to the service/endpoints to retrieve those metrics?

So we need to discuss this - these metrics were intended for operations staff and not (necessarily) to be exposed to end users. I do think however we will want to use some of these metrics part of the product (users should be able to see the rate of data coming in per agent, for example), so we need to discuss making sure the Prometheus used for this is part of the production system. cc @stefano-daitan @iatdaitan

weyrick (approved):
Great, thanks for working through this. Let's make sure we make a backlog ticket for the "double identify" tech debt

dscabral (approved):

PR #530, opened 2021-11-16
fix(orb-ui): unable edit {entity} no tags

gpazuch:
resolves #512
resolves #191 
resolves #183
resolves #366

PR #532, opened 2021-11-17
fix(orb-ui): agent group tag issues

### LINKED ISSUE #183: Agent Groups: Although an error message appears, it is possible to create two identical key and value pairs
A error message is being displayed but double key-value creation is allowed

Check: https://drive.google.com/file/d/1MvdknGxMO03T9Npfqo35hUCbqvXp44ib/view?usp=sharing

Last Commit: 50df003b85d894ec2f3ba7d5d7a913eec1e82e4e

### LINKED ISSUE #366: Agent Groups: When all tags are removed, agent match is not updated
When an agent group is created or edited, if tag(s) are included and then all are removed, agent match is not updated

gpazuch:
resolves #183 
resolves #366


gpazuch:
@iatdaitan this PR branches of #530 , it should only be merged afterwards completion of #530

PR #533, opened 2021-11-17
fix(orb-ui): #485 'save' instead of 'next'

### LINKED ISSUE #485: Policy Management: create and edit policy is without "save" option and is saving by "next" option
Like all other components, it is expected that when a policy is created or any existing policy is edited, the last page will be a review and confirmation page, with the option to go back and save, but the policy is being saved by clicking " next" button without going through review and confirm. Check:

gpazuch:
resolves #485


iatdaitan (commented):
lgtm

iatdaitan (approved):

PR #559, opened 2021-11-19
Python test

manrodrigues:


gpazuch (commented):
Is it necessary to commit `python-test/features/stesp/__pycache__/functions.cpython-38.pyc` binary?

python-test/requirements.txt
```diff
@@ -0,0 +1,90 @@
+argon2-cffi==21.1.0
```
weyrick, line 1: is all of this actually used? would be great to limit this to exactly what we need to keep CI pipelines fast

python-test/features/steps/agentsProvider.py
```diff
@@ -0,0 +1,55 @@
+from behave import given, when, then
+import time
+import functions
+
+base_orb_url = "https://beta.orb.live"
```
weyrick, line 5: I'm not sure how variables work in behave, but it seems like this should be pulled out somewhere?

python-test/features/steps/agentsProvider.py
```diff
@@ -0,0 +1,55 @@
...
+    context.status_agent, context.agent_data = functions.new_agent(context.token, agentname, agentTagKey, agentTagValeu)
+    assert context.status_agent == 201
+    context.agent_key = context.agent_data['key']
+    context.agent_id = context.agent_data['id']
+    context.agent_channel_id = context.agent_data['channel_id']
+
+@when('Build agent container')
+def build_agent(context):
+    iface = 'mock'
+    provision_agent_command = f"docker run -d --net=host -e ORB_CLOUD_ADDRESS=beta.orb.live -e ORB_CLOUD_MQTT_ID={context.agent_id} -e ORB_CLOUD_MQTT_CHANNEL_ID={context.agent_channel_id} -e ORB_CLOUD_MQTT_KEY={context.agent_key} -e PKTVISOR_PCAP_IFACE_DEFAULT={iface} ns1labs/orb-agent"
```
weyrick, line 30: the way you have it, this will use the `latest` tag for `orb-agent` - this probably needs to be variable too, and often in our development pipelines it needs to be set to `develop` so that it reads: `ns1labs/orb-agent:develop`

here we would use a variable for "beta.orb.live" too

python-test/features/steps/agentsProvider.py
```diff
@@ -0,0 +1,55 @@
...
+
+@when('Create an agent')
+def create_agent(context):
+    context.status_agent, context.agent_data = functions.new_agent(context.token, agentname, agentTagKey, agentTagValeu)
+    assert context.status_agent == 201
+    context.agent_key = context.agent_data['key']
+    context.agent_id = context.agent_data['id']
+    context.agent_channel_id = context.agent_data['channel_id']
+
+@when('Build agent container')
```
weyrick, line 27: this should say "Run agent container", because "building" a container is a different operation

python-test/features/steps/agentsProvider.py
```diff
@@ -0,0 +1,55 @@
...
+def build_agent(context):
+    iface = 'mock'
+    provision_agent_command = f"docker run -d --net=host -e ORB_CLOUD_ADDRESS=beta.orb.live -e ORB_CLOUD_MQTT_ID={context.agent_id} -e ORB_CLOUD_MQTT_CHANNEL_ID={context.agent_channel_id} -e ORB_CLOUD_MQTT_KEY={context.agent_key} -e PKTVISOR_PCAP_IFACE_DEFAULT={iface} ns1labs/orb-agent"
+    context.id_terminal = functions.send_terminal_commands(provision_agent_command, '[')
+    assert len(context.id_terminal) > 0 
+
+@then("Agente should be online")
```
weyrick, line 34: typo on "Agent"

python-test/features/steps/functions.py
```diff
@@ -0,0 +1,123 @@
...
+        delete_agent = requests.delete(base_orb_url+'/api/v1/agents/'+list_of_agents[agent]['id'], headers={'Authorization': token})
+        assert delete_agent.status_code == 204
+
+def new_agent(token, name, tagKey, tagValue, base_orb_url = base_orb_url):
+    new_agent = requests.post(base_orb_url+'/api/v1/agents', json= {"name": name,"orb_tags": {tagKey: tagValue}, "validate_only": False}, headers={'Content-type':'application/json',  'Accept':'*/*', 'Authorization': token})
+    return new_agent.status_code, json.loads(new_agent.text)
+
+def send_terminal_commands(command, separator=None, cwd_run=None):
+    args = shlex.split(command)
+    docker_command_execute = subprocess.Popen(args, stdout=subprocess.PIPE, cwd=cwd_run)
```
weyrick, line 51: Is the long term strategy to use a terminal subprocess instead of interacting with docker directly?

python-test/features/steps/functions.py
```diff
@@ -0,0 +1,123 @@
...
+    if separator == None:
+        subprocess_return_terminal = subprocess_return.split()
+    else:
+        subprocess_return = subprocess_return.split()
+        subprocess_return_terminal = subprocess_return[0].split(separator)
+    return subprocess_return_terminal[0]
+
+def check_logs(logs):
+    match = False
+    providing_log_regex = "'level':'info','ts':.+,'caller':'agent/rpc_to.go:.+','msg':'sending capabilities','value':'{.'schema_version.':.+,.'orb_agent.':{.'version.':.+},.'agent_tags.'.+.'backends.':{.'pktvisor.':{.'version.':.+.',.'data.':{.'taps.':{.'default_pcap.':{.'config.':{.'iface.':.+,.'pcap_source.':.+},.'input_type.':.'pcap.',.'interface.':.+.'"
```
weyrick, line 62: Unfortunately this is fragile - this schema is not guaranteed. This log output though was meant to be parsed as normal JSON, precisely for automation like this so let's discuss how we can get that in place instead of text matching

python-test/features/steps/agentsProvider.py
```diff
@@ -0,0 +1,55 @@
...
+@when('Create an agent')
+def create_agent(context):
+    context.status_agent, context.agent_data = functions.new_agent(context.token, agentname, agentTagKey, agentTagValeu)
+    assert context.status_agent == 201
+    context.agent_key = context.agent_data['key']
+    context.agent_id = context.agent_data['id']
+    context.agent_channel_id = context.agent_data['channel_id']
+
+@when('Build agent container')
+def build_agent(context):
```
weyrick, line 28: run_agent

python-test/features/steps/functions.py
```diff
@@ -0,0 +1,123 @@
...
+            i_replaced = i.replace('"',"'")
+            print(str(i_replaced))
+            match_agent = re.findall(providing_log_regex, str(i_replaced))
+            if match_agent:
+                match = True
+    except ValueError:
+        print("Invalid container ID")
+    return match
+
+def agent_provising(agentname, tagKey, tagValue, token, iface="mock"):
```
weyrick, line 74: I'm unclear on why this is repeated here

weyrick (changes requested):
Good start! Please see comments. Also, we should add ".pyc" to .gitignore, as these cache files should not be part of the PR

python-test/features/steps/functions.py
weyrick: The way I would prefer to do it would be direct calls to docker with https://docker-py.readthedocs.io/en/stable/

python-test/features/steps/functions.py
manrodrigues: it shouldn't. I was using like this, but i preferred to separate and i forgot to erase. I'm gonna remove it.

python-test/requirements.txt
manrodrigues: I'll try to filter only the extremely necessary

python-test/features/steps/agentsProvider.py
manrodrigues: you are right. Maybe as an environment variable, just like email and password. I'll talk to Stefano about the best way to do it.

python-test/features/steps/agentsProvider.py
manrodrigues: Thanks. I'm gonna change it

stefano-daitan:
Also, create a template file with all these environment variables, add the 'non-template' to .gitignore, and document how to run these tests (maybe a README.md in `python-test/` dir?.

python-test/features/steps/functions.py
```diff
@@ -0,0 +1,77 @@
+import requests
+import json
+# from pygments import highlight, lexers, formatters
```
stefano-daitan, line 3: remove unused comment

python-test/features/steps/functions.py
```diff
@@ -0,0 +1,77 @@
...
+import os
+import docker
+import re
+
+base_orb_url = os.getenv('ORB_URL', "https://beta.orb.live")
```
stefano-daitan, line 8: why do we need two variables, `ORB_ADDRESS` and `ORB_URL`?
We could have a single `ORB_ADDRESS` and do
```python
base_orb_url = "https://" + os.getenv('ORB_ADDRESS', "beta.orb.live")

File: python-test/features/steps/functions.py [hunk 12]
```diff
@@ -0,0 +1,77 @@
...
+
+
+
+def check_existing_agents(token, id= "" , base_orb_url = base_orb_url):
+    agentDict = dict()
+    check_agents = requests.get(base_orb_url+'/api/v1/agents'+id, headers={'Authorization': token})
+    agents_as_json = json.loads(check_agents.text)
+    formatted_json = json.dumps(agents_as_json, sort_keys=True, indent=4)
+    # colorful_json = highlight(formatted_json, lexers.JsonLexer(), formatters.TerminalFormatter())
+    # print(colorful_json)
```
stefano-daitan, line 32: remove comments

python-test/features/steps/agentsProvider.py
```diff
@@ -0,0 +1,54 @@
...
+import functions
+import os
+import docker
+
+email = os.getenv('EMAIL')
+password = os.getenv('PASSWORD')
+
+agentname = os.getenv('AGENT_NAME')
+agentTagKey = os.getenv('TAG_KEY', 'test')
+agentTagValeu = os.getenv('TAG_VALUE', 'true')
```
stefano-daitan, line 12: typo: agentTagVal**eu** -> agentTagValue

python-test/features/steps/agentsProvider.py
```diff
@@ -0,0 +1,54 @@
...
+    context.token = functions.generate_token(email, password)
+    assert len(context.token) > 0
+
+@when('Create an agent')
+def create_agent(context):
+    context.status_agent, context.agent_data = functions.new_agent(context.token, agentname, agentTagKey, agentTagValeu)
+    assert context.status_agent == 201
+    context.agent_key = context.agent_data['key']
+    context.agent_id = context.agent_data['id']
+    context.agent_channel_id = context.agent_data['channel_id']
```
stefano-daitan, line 26: This test is not idempotent. Running it twice fails and we are required to manually delete the agent on Orb UI.
How about deleting the agent at the end of the test?

python-test/features/steps/agentsProvider.py
```diff
@@ -0,0 +1,54 @@
+from behave import given, when, then
+import time
+import functions
+import os
+import docker
+
+email = os.getenv('EMAIL')
+password = os.getenv('PASSWORD')
+
+agentname = os.getenv('AGENT_NAME')
```
stefano-daitan, line 10: Let's have a default name for the agent (maybe a random name?).
When running the tests, we must provide flexibility (like having this env variable) but we should also hide as much complexity as we can.

python-test/features/steps/agentsProvider.py
```diff
@@ -0,0 +1,54 @@
...
+agentTagKey = os.getenv('TAG_KEY', 'test')
+agentTagValeu = os.getenv('TAG_VALUE', 'true')
+
+
+@given('A valid authentication')
+def get_auth(context):
+    context.token = functions.generate_token(email, password)
+    assert len(context.token) > 0
```
stefano-daitan, line 18: We should use assertion libraries that provide useful functionalities like nice error messages, comparisons beyond boolean, and improved readability.

python-test/features/steps/agentsProvider.py
```diff
@@ -0,0 +1,54 @@
...
+    context.agent_key = context.agent_data['key']
+    context.agent_id = context.agent_data['id']
+    context.agent_channel_id = context.agent_data['channel_id']
+
+@when('Run agent container')
+def run_agent(context):
+    iface = os.getenv('IFACE', 'mock')
+    context.orb_address = os.getenv('ORB_ADDRESS', 'beta.orb.live')
+    context.version_image = os.getenv('AGENT_VERSION')
+    context.agent_image = f"ns1labs/orb-agent:{context.version_image}"
```
stefano-daitan, line 33: We shouldn't include the version field in the image if it was not provided.
In that case, `agent_image` would be simply `"ns1labs/orb-agent"` without the `:`.

python-test/features/steps/functions.py
```diff
@@ -0,0 +1,77 @@
...
+
+base_orb_url = os.getenv('ORB_URL', "https://beta.orb.live")
+
+
+def generate_token(email, password, base_orb_url = base_orb_url):
+    if email != None and password != None: 
+        headers={'Content-type':'application/json',  'Accept':'*/*'}
+        token_request = requests.post(base_orb_url+'/api/v1/tokens', json= {'email': email, 'password': password}, headers=headers)
+        if token_request.status_code >= 200 and token_request.status_code <300:
+            return (((token_request.text).split(":"))[1]).split('"')[1]
```
stefano-daitan, line 16: Is the response a JSON string? Should we parse it instead of manipulating the string?

python-test/features/steps/agentsProvider.py
```diff
@@ -0,0 +1,54 @@
...
+    var_env ={"ORB_CLOUD_ADDRESS":context.orb_address, "ORB_CLOUD_MQTT_ID":context.agent_id,"ORB_CLOUD_MQTT_CHANNEL_ID":context.agent_channel_id, "ORB_CLOUD_MQTT_KEY":context.agent_key,"PKTVISOR_PCAP_IFACE_DEFAULT":iface}
+    client = docker.from_env()
+    context.agent_container = client.containers.run(context.agent_image, detach=True, network_mode= 'host', environment=var_env)
+    assert len(context.agent_container.id) > 0
+    context.id_terminal =  context.agent_container.id
```
stefano-daitan, line 38: We should delete the container after the test is done.

python-test/features/steps/agentsProvider.py
```diff
@@ -24,14 +25,19 @@ def create_agent(context):
...
-    provision_agent_command = f"docker run -d --net=host -e ORB_CLOUD_ADDRESS=beta.orb.live -e ORB_CLOUD_MQTT_ID={context.agent_id} -e ORB_CLOUD_MQTT_CHANNEL_ID={context.agent_channel_id} -e ORB_CLOUD_MQTT_KEY={context.agent_key} -e PKTVISOR_PCAP_IFACE_DEFAULT={iface} ns1labs/orb-agent"
-    context.id_terminal = functions.send_terminal_commands(provision_agent_command, '[')
-    assert len(context.id_terminal) > 0 
-
-@then("Agente should be online")
+@when('Run agent container')
+def run_agent(context):
+    iface = os.getenv('IFACE', 'mock')
+    context.orb_address = os.getenv('ORB_ADDRESS', 'beta.orb.live')
+    context.version_image = os.getenv('AGENT_VERSION')
```
weyrick, line 32: let's make this `AGENT_DOCKER_TAG` instead of `AGENT_VERSION`, because they are not quite the same thing: the tag may be `develop` for example which could be any version.

python-test/features/steps/agentsProvider.py
weyrick: I think behave has a canonical way to do this type of cleanup, we should use whatever it is.

python-test/features/steps/functions.py
weyrick: agreed we only need one variable here. but note that this variable is currently processed by the docker entry script, not orb-agent directly

https://github.com/ns1labs/orb/blob/develop/agent/docker/orb-agent-entry.sh#L10

python-test/features/steps/functions.py
weyrick: agreed. `requests` library makes this easy, just use `token_request.json()` instead of `.text`

https://docs.python-requests.org/en/latest/user/quickstart/#json-response-content

python-test/features/steps/functions.py
```diff
@@ -1,32 +1,34 @@
...
+        assert email != None
+        assert password != None
+
+
 
 
 def check_existing_agents(token, id= "" , base_orb_url = base_orb_url):
     agentDict = dict()
     check_agents = requests.get(base_orb_url+'/api/v1/agents'+id, headers={'Authorization': token})
     agents_as_json = json.loads(check_agents.text)
```
weyrick, line 29: again you can use `.json()` here intest of `.text`

python-test/features/steps/functions.py
```diff
@@ -46,46 +48,20 @@ def new_agent(token, name, tagKey, tagValue, base_orb_url = base_orb_url):
...
-        subprocess_return = subprocess_return.split()
-        subprocess_return_terminal = subprocess_return[0].split(separator)
-    return subprocess_return_terminal[0]
 
 def check_logs(logs):
     match = False
     providing_log_regex = "'level':'info','ts':.+,'caller':'agent/rpc_to.go:.+','msg':'sending capabilities','value':'{.'schema_version.':.+,.'orb_agent.':{.'version.':.+},.'agent_tags.'.+.'backends.':{.'pktvisor.':{.'version.':.+.',.'data.':{.'taps.':{.'default_pcap.':{.'config.':{.'iface.':.+,.'pcap_source.':.+},.'input_type.':.'pcap.',.'interface.':.+.'"
     try:
         for i in logs:
             i_replaced = i.replace('"',"'")
```
weyrick, line 57: I'm pretty sure `i` should be valid json. Can you try `logline = json.loads(i)` and see if we can use that interface instead of text matching? Then you can search for match on `logline['msg'] == 'sending capabilities'` for example?

weyrick (changes requested):

python-test/features/steps/agentsProvider.py
stefano-daitan: I don't think that will be a problem, right @weyrick ?

weyrick: Well it's not a problem per se, but I'm just wondering if we will get the results we expect. The tag for `orb-agent` should probably match the tag/version of the environment we are testing against

weyrick: `latest` will always match the latest version from self-host release, so 0.9.0 right now. As `develop` progresses, it may have had changes that we need for the tests to succeed.

python-test/features/steps/functions.py
```diff
@@ -0,0 +1,84 @@
...
+
+def generate_token(email, password, base_orb_url = base_orb_url):
+    if email != None and password != None: 
+        headers={'Content-type':'application/json',  'Accept':'*/*'}
+        token_request = requests.post(base_orb_url+'/api/v1/tokens', json= {'email': email, 'password': password}, headers=headers)
+        hamcrest.assert_that(token_request.json(), hamcrest.has_key('token'), 'Request Json - token')
+        return token_request.json()
+    else:
+        hamcrest.assert_that(email, hamcrest.not_none(), 'Email')
+        hamcrest.assert_that(password, hamcrest.not_none(), 'Password')
```
stefano-daitan, line 18: Why not asserting before the if/else block? In fact, if we do that we won't need the if/else in the first place.

python-test/features/steps/functions.py
```diff
@@ -0,0 +1,84 @@
...
+import hamcrest
+
+base_orb_url = "https://" + os.getenv('ORB_ADDRESS', "beta.orb.live")
+
+def generate_token(email, password, base_orb_url = base_orb_url):
+    if email != None and password != None: 
+        headers={'Content-type':'application/json',  'Accept':'*/*'}
+        token_request = requests.post(base_orb_url+'/api/v1/tokens', json= {'email': email, 'password': password}, headers=headers)
+        hamcrest.assert_that(token_request.json(), hamcrest.has_key('token'), 'Request Json - token')
+        return token_request.json()
```
stefano-daitan, line 15: Don't we need to extract the token value from the response json?

python-test/README.md
```diff
@@ -0,0 +1,40 @@
...
+
+`$ source name_of_virtualenv/bin/activate`
+
+## Install the required libraries
+
+`$ pip install -r requirements.txt`
+
+
+## Configuring the correct variables
+Modified the ./steps/env_variables.env file changing the correct values
```
stefano-daitan, line 19: Modi**fy**

python-test/features/steps/local_agent.py
```diff
@@ -0,0 +1,90 @@
...
+
+configs = TestConfig.configs()
+
+
+@when('the agent container is started')
+def run_local_agent_container(context):
+    orb_address = configs.get('orb_address')
+    interface = configs.get('orb_agent_interface', 'mock')
+    image_tag = ':' + configs.get('agent_docker_tag', 'latest')
+    agent_image = "ns1labs/orb-agent" + image_tag
```
weyrick, line 15: Occurred to me we should make "ns1labs" a config variable as well - a community member may wish to use images from their own repository

weyrick (approved):
Looking great! Very clean now. Just a note that we should make "ns1labs" repo optionally configurable.

stefano-daitan (approved):

PR #562, opened 2021-11-22
Refact/orb dataset agent architecture

### LINKED ISSUE #449: refactor agent dataset/policy RPC architecture
* keep complexity of dataset ID management out of the agent
* send only AgentPolicyRPC, remove dataset ids (?)
* remove DatasetRemovedRPC
* requires change in how we determine how to sink policy metrics in sinker
* probably requires us to set new minimum orb-agent version

references:
* https://github.com/ns1labs/orb/wiki/Architecture:-Fleet#core-to-agent-communication-paths-for-synchronized-state
* #414 
* https://ns1labs.slack.com/archives/C01UNG6UK37/p1636059648271600

dscabral:
Resolves #449


go.mod
```diff
@@ -5,6 +5,7 @@ go 1.15
 require (
 	github.com/eclipse/paho.mqtt.golang v1.2.0
 	github.com/fatih/structs v1.1.0
+	github.com/fullstorydev/grpcurl v1.8.5 // indirect
```
dscabral, line 8: Remove it before PR

buildinfo/version.go
```diff
@@ -14,7 +14,7 @@ import (
 )
 
 // set via ldflags -X option at build time
-var version = "unknown"
+var version = "0.9.0-develop"
```
dscabral, line 17: Version should be "unknown"

policies/pb/policies.proto
```diff
@@ -49,5 +50,10 @@ message DatasetRes {
   string id = 1;
   string agent_group_id = 2;
   string policy_id = 3;
-  repeated string sink_ids = 4;
+  bool valid = 4;
```
weyrick, line 53: now that we will have our first users and production environments, we need to start thinking about backwards compatibility here - I'm pretty sure you should never change a field's number, you can only add new ones. we should read up on this

https://earthly.dev/blog/backward-and-forward-compatibility/
https://developers.google.com/protocol-buffers/docs/proto3#assigning_field_numbers

weyrick (changes requested):
Looks pretty good but I realized one important thing here: we still need to be producing and consuming dataset removed redis events, so we need to leave that code in. However, now instead of the fleet consumer calling `NotifyGroupDatasetRemoval`, it's got to run some logic to see if it should send `AgentPolicyRPCPayload` with "remove" action to tell the agent(s) to stop running that policy. Essentially we have moved the complexity from the agents into this one function. If we don't do this, there is no way an agent would know to stop running a policy when a dataset is deleted. However, since there maybe overlap between policies and datasets, we can't just send `AgentPolicyRPCPayload` "remove" until we check that no other datasets want that policy to run. Let's discuss.

weyrick:
I recommend we close this for now. You can save the branch if you want for reference though, when we get around to looking at this problem again

PR #566, opened 2021-11-26
doc(fleet): update fleet service openapi spec

### LINKED ISSUE #534: API Responses are wrong or outdated on API Docs home page
CreateAgent service response is wrong or outdated under the API's Docs home page. For example, https://getorb.io/api/fleet.html#operation/createAgent should also return the Agent Key and Channel ID.

The openapi yaml file need to be updated to reflect the right response.

mclcavalcante:
Resolves #534

- add key and channel id to Agent object schema
- add response schema of agent group validate

### DIFF
--- a/fleet/api/http/openapi.yaml
+++ b/fleet/api/http/openapi.yaml
@@ -115,7 +115,7 @@ paths:
         $ref: "#/components/requestBodies/AgentGroupCreateReq"
       responses:
         '200':
-          description: Valid Agent group
+          $ref: "#/components/responses/AgentGroupValidateObjRes"
         '400':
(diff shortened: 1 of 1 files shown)


fleet/api/http/openapi.yaml
```diff
@@ -705,6 +734,14 @@ components:
           type: string
           format: date-time
           description: A time stamp of the last heartbeat that was received
+        key:
+          type: string
+          format: uuid
+          description: Agent key is a secret key that's generated at the agent creation for authentication purposes
```
weyrick, line 740: small grammar correction "generated at agent creation"

PR #581, opened 2021-12-02
feat(orb-ui): integrate pactsafe clickwrapper

gpazuch:
Should display contract on register form.

CSS for contract needs minor adjustments - height overflows parent container.


gpazuch:
@iatdaitan so, there are a couple issues with css.

it should run ok on localhost as long as you set the .env file at ui/.env
with
```# IronClad Conf. Params.
PS_SID="theside id"
PS_GROUP_KEY="thegroupkey"```

gpazuch:
before clickbox
after clickbox
contract page view

gpazuch:
@weyrick I just posted a few screenshots of current implementation.
Removed the contract display, only checkbox visible, contract opens to new link.

There is a bit of work left to do still, which is to keep register button disabled while form is non valid || contract isn't accepted.

weyrick:
> @weyrick I just posted a few screenshots of current implementation. Removed the contract display, only checkbox visible, contract opens to new link.
> 
> There is a bit of work left to do still, which is to keep register button disabled while form is non valid || contract isn't accepted.

Ok let's land the basic flow even if we need to make it prettier later.

iatdaitan (approved):
Approved.

Going to add an enhancement issue to improve the UX/UI.

PR #601, opened 2021-12-09
Feat/orb otelco

dscabral:
Create a prom receiver on agent start up to scrapt from http://localhost:10853/
Create a simple prom exporter to show on http://localhost:8787/metrics


cmd/agent/main.go
```diff
@@ -8,24 +8,28 @@ import (
...
 const (
 	defaultConfig = "/etc/orb/agent.yaml"
 )
 
 var (
-	cfgFiles []string
-	Debug    bool
+	cfgFiles                  []string
+	Debug                     bool
+	rng                       = rand.New(rand.NewSource(time.Now().UnixNano()))
```
weyrick, line 31: these changes can go away right?

go.mod
```diff
@@ -8,29 +8,39 @@ require (
...
 	github.com/go-cmd/cmd v1.3.0
 	github.com/go-co-op/gocron v1.9.0
-	github.com/go-kit/kit v0.10.0
+	github.com/go-kit/kit v0.11.0
 	github.com/go-redis/redis/v8 v8.11.0
 	github.com/go-zoo/bone v1.3.0
-	github.com/gofrs/uuid v3.3.0+incompatible
-	github.com/golang/protobuf v1.5.0
-	github.com/golang/snappy v0.0.1
+	github.com/gofrs/uuid v4.0.0+incompatible
```
weyrick, line 14: this is a major version upgrade, we should check the release notes to make sure this will not break anything in the other orb services

go.mod
```diff
@@ -8,29 +8,39 @@ require (
...
+	github.com/lib/pq v1.9.0
 	github.com/mainflux/mainflux v0.12.0
 	github.com/mattn/go-sqlite3 v1.14.6
 	github.com/mitchellh/mapstructure v1.4.2
 	github.com/opentracing/opentracing-go v1.2.0
 	github.com/ory/dockertest/v3 v3.6.0
-	github.com/prometheus/client_golang v1.7.1
-	github.com/prometheus/prometheus v2.5.0+incompatible
+	github.com/prometheus/client_golang v1.11.0
+	github.com/prometheus/prometheus v1.8.2-0.20210621150501-ff58416a0b02
```
weyrick, line 26: this is actually a _downgrade_, any idea why? does it break our orb-sinker?

go.mod
```diff
@@ -8,29 +8,39 @@ require (
...
-	go.uber.org/zap v1.16.0
-	google.golang.org/grpc v1.38.0
+	github.com/uber/jaeger-client-go v2.29.1+incompatible
+	go.uber.org/zap v1.19.1
+	google.golang.org/grpc v1.42.0
 	google.golang.org/protobuf v1.27.1
 	gopkg.in/yaml.v3 v3.0.0-20210107192922-496545a6307b
 )
+
+require (
```
weyrick, line 38: why is this a separate require section - is this the direct requires for new otel stuff? if so and we want it to be a separate section, let's comment that.

sinker/prometheus/client.go
```diff
@@ -264,20 +264,20 @@ func (c *client) WriteProto(
 
 // toPromWriteRequest converts a list of timeseries to a Prometheus proto write request.
 func (t TSList) toPromWriteRequest() *prompb.WriteRequest {
-	promTS := make([]*prompb.TimeSeries, len(t))
+	promTS := make([]prompb.TimeSeries, len(t))
```
weyrick, line 267: curious why this change was made?

agent/backend/pktvisor/pktvisor.go
```diff
@@ -258,6 +269,29 @@ func (p *pktvisorBackend) Start() error {
 	p.scraper = gocron.NewScheduler(time.UTC)
 	p.scraper.StartAsync()
 
+	//TODO create a startup for otel receiver / exporter component
```
weyrick, line 272: we discussed this needing to be featured flagged with a new configuration item in orb config, with a default to off. that way we can make these 0.10.0 release without using it by default and breaking any existing functionality

weyrick (changes requested):

agent/backend/pktvisor/pktvisor.go
```diff
@@ -269,93 +272,94 @@ func (p *pktvisorBackend) Start() error {
...
-	if err != nil {
-		p.logger.Error("otel receiver startup error", zap.Error(err))
-		os.Exit(1)
-	}
-
-	// scrape all policy json output with one call every minute.
-	// TODO support policies with custom bucket times
-	job, err := p.scraper.Every(1).Minute().WaitForSchedule().Do(func() {
-		metrics, err := p.scrapeMetrics(1)
+	if p.scrapeOtel {
```
weyrick, line 275: this is large enough to be hard to read now, please split each fork in this logic block here (setting up otel and setting up the other scraper) out to its own pktvisorBackend method

agent/config/types.go
```diff
@@ -35,12 +35,17 @@ type Cloud struct {
 	MQTT   MQTTConfig  `mapstructure:"mqtt"`
 }
 
+type Opentelemetry struct {
+	Scrape bool `mapstructure:"scrape"`
```
weyrick, line 39: Sorry I should have been more specific on the naming - can you name this "Enabled" instead of "Scrape" and adjust the rest of the code changes here to change that word to please?

weyrick (changes requested):

agent/agent.go
```diff
@@ -75,7 +75,7 @@ func (a *orbAgent) startBackends() error {
 			return errors.New("specified backend does not exist: " + name)
 		}
 		be := backend.GetBackend(name)
-		config["scrape_otel"] = strconv.FormatBool(a.config.OrbAgent.Otel.Scrape)
+		config["scrape_otel"] = strconv.FormatBool(a.config.OrbAgent.Otel.Enable)
```
weyrick, line 78: I didn't notice the placement of this before - this is not very clean. I would rather add another parameter to Backend.Configure() interface and pass in the whole Otel config (after the backend config, like `Configure(*zap.Logger, policies.PolicyRepo, map[string]string, map[string]string) error` or move this config from `orb.otel.enable` to `orb.backends.pktvisor.otel.enable` so that it's in the existing config structure

agent/backend/pktvisor/pktvisor.go
```diff
@@ -273,96 +273,108 @@ func (p *pktvisorBackend) Start() error {
...
+	}
+	p.receiver, err = createReceiver(ctx, p.exporter, p.logger)
+	if err != nil {
+		p.logger.Error("failed to create a receiver", zap.Error(err))
+	}
+
+	err = p.exporter.Start(ctx, nil)
+	if err != nil {
+		p.logger.Error("otel exporter startup error", zap.Error(err))
+		os.Exit(1)
```
weyrick, line 368: let's get rid of these `os.Exit()` calls and these `p.logger.Error` messages and instead do a `return err` the same as `scrapeDefault`, and let the agent startup handle the error

weyrick (changes requested):
Almost there, just a few small items please

weyrick (approved):

Issue #362, opened 2021-10-06
Incomplete reload pages after delete


manrodrigues:
When the last existing Sink, Agent or Agent Group is deleted, page should be reload and display "No data to display", but the expected message is not being displayed;

manrodrigues:
![nodata](https://user-images.githubusercontent.com/78241475/136226862-07b0a1ed-ea06-4615-aa05-511f4e140bd7.gif)

manrodrigues:
This is still not working <br> 

![reload_after_delete](https://user-images.githubusercontent.com/78241475/146414996-95a71b5d-afb4-4934-9d12-8067dabef727.gif)

manrodrigues:
For agent groups, datasets, policies and sinks... its working fine

manrodrigues:
![agent-delete-ok](https://user-images.githubusercontent.com/78241475/146602052-2478fd2c-9c02-4818-a25f-443a96f105b7.gif)

working fine for agents as well

manrodrigues:
@gpazuch i'm reopening this issue, because we broken the reload of dataset's page and for groups page the loading bar is being displayed after removing.. so is incomplete:

![refresh_groups](https://user-images.githubusercontent.com/78241475/161982952-2367bbb4-6860-4412-80cf-8871d97a39b0.gif)

![refresh_dataset](https://user-images.githubusercontent.com/78241475/161982990-99dde0c6-0494-4d42-9374-3ddd9d34e5b8.gif)

manrodrigues:
ok. LGTM


![group_remove](https://user-images.githubusercontent.com/78241475/163865507-c42c369c-3390-476d-ba8d-52e2838b23d2.gif)


![dataset_remove](https://user-images.githubusercontent.com/78241475/163865513-4c130de5-8137-402d-b3b2-ca6810de347f.gif)

PR #605, opened 2021-12-13
fix(orb-ui): #362 UI reload incomplete

gpazuch:
fix(orb-ui): pages never stop reloading
* fix for agent groups list page
* fix for agents list page
* fix for datasets list page
* fix for agent policies list page


iatdaitan (approved):
Well done.

PR #618, opened 2021-12-15
feat(policies): add service metrics to api

### LINKED ISSUE #607: Instrumentation for Policies Service Metrics
- request latency
- number of requests

mclcavalcante:


dscabral (approved):
Looks pretty good... but I'm not sure if using policy_id and dataset_id together it's the best approach... Maybe we should segregate each one into a new service (dataset service and policy service) than it will be possible to instrumentate each one.

weyrick (approved):

Issue #625, opened 2021-12-16
Reorder home information blocks to match documentation


iatdaitan:
Although the user is allowed to follow a different path during the use of Orb, it makes sense to show the information as it is stated on the documentation to increase UX.

This issue aims to match the home information blocks according to the path followed on the Orb Documentation.

![image](https://user-images.githubusercontent.com/86745196/146374226-2dd59cc3-c2d4-4dc3-8dbc-c85c710196ea.png)

- Agents
- Agent Groups
- Sink 
- Policy
- Datasets

PR #637, opened 2021-12-16
fix(orb-ui): #625 fix order on dashboard page

gpazuch:
#625 Fix order of items on dashboard page


gpazuch:
There are still a couple adjustments, I'll push soon

manrodrigues (approved):

PR #664, opened 2021-12-21
fix missing PKTVISOR_TAG in GitHub actions. bring go builds up to 1.17

weyrick:


dscabral (approved):
It's working locally for me

PR #667, opened 2021-12-21
testing agents through the ui

manrodrigues:


python-test/features/steps/local_agent.py
```diff
@@ -15,11 +16,20 @@ def run_local_agent_container(context):
...
-                "ORB_CLOUD_MQTT_KEY": context.agent['key'],
-                "PKTVISOR_PCAP_IFACE_DEFAULT": interface}
+    if "localhost" in base_orb_url:
+        env_vars = {"ORB_CLOUD_ADDRESS": 'localhost',
+                    "ORB_CLOUD_MQTT_ID": context.agent['id'],
+                    "ORB_CLOUD_MQTT_CHANNEL_ID": context.agent['channel_id'],
+                    "ORB_CLOUD_MQTT_KEY": context.agent['key'],
+                    "PKTVISOR_PCAP_IFACE_DEFAULT": interface,
+                    "ORB_TLS_VERIFY": "false"}
+    else:
```
stefano-daitan, line 26: As far as I can tell, this `if` block is only different than the `else` block because of the `ORB_TLS_VERIFY` field.
In that case, let's add a config parameter for TLS verification rather than branching the code.
Question: why aren't HTTP calls also affected by this self-signed SSL certificates?

python-test/features/steps/login_ui.py
```diff
@@ -0,0 +1,74 @@
...
+user_email = configs.get('email')
+user_password = configs.get('password')
+
+
+@given("that the user is on the orb page")
+def orb_page(context):
+    """Open the required page on Chrome browser
+    Args:
+        context (behave.runner.Context): object that hold contextual information during the running of tests
+    """
```
stefano-daitan, line 19: No need to add docs in step functions. The `@given` should be clear enough so we understand what this function does.

python-test/features/steps/login_ui.py
```diff
@@ -0,0 +1,74 @@
...
+
+def go_to_page(page, context):
+    """Open the page in Chrome browser
+    Args:
+        page (string): site's URL
+        context (behave.runner.Context): object that hold contextual information during the running of tests
+    """
+    options = webdriver.ChromeOptions()
+    options.add_argument('--ignore-ssl-errors=yes')
+    options.add_argument('--ignore-certificate-errors')
```
stefano-daitan, line 48: Ignoring SSL/certificate errors should be configurable.

python-test/features/steps/login_ui.py
```diff
@@ -0,0 +1,74 @@
...
+    options.add_argument("--start-maximized")
+    context.driver = webdriver.Chrome(options=options)
+    context.driver.get(str(page))
+    assert_that(context.driver.current_url, equal_to(f"{base_orb_url}/auth/login"), "user not enabled to access orb "
+                                                                                    "login page")
+
+
+def enter_information(html_information_name, information, context, source_by=0):
```
stefano-daitan, line 56: Looks like `enter_information` and `go_to_page` can be used in many use-cases, not only in login.
Let's create a `ui_utils.py` module and add all functions that are useful to multiple UI-related scenarios.

Also, let's stick with UI nomenclature. So `input_text` is more suitable than `enter_information`, what do you think?

python-test/features/steps/login_ui.py
```diff
@@ -0,0 +1,74 @@
...
+        data = context.driver.find_element_by_id(str(html_information_name))
+        data.send_keys(str(information))
+    elif source_by == 1:
+        data = context.driver.find_element_by_css_selector(str(html_information_name))
+        data.send_keys(str(information))
+    elif source_by == 2:
+        data = context.driver.find_element_by_xpath(str(html_information_name))
+        data.send_keys(str(information))
+    else:
+        raise Exception("source_by not found")
```
stefano-daitan, line 74: This sort of if/else with numeric values representing different element types is not a good way of inputting the text correctly. Let's think of something better.

python-test/features/steps/users.py
```diff
@@ -32,3 +43,20 @@ def authenticate(user_email, user_password):
     assert_that(response.json(), has_key('token'), 'Authentication token not found in login response!')
 
     return response.json()['token']
+
+
+def register_account(user_email, user_password, request_status):
```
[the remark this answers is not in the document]
manrodrigues: if I encapsulate the result I can only use the function to register valid accounts. I did so thinking about reusing it for future tests, with unfortunate paths. Should I make a function for each type of "assert"?

python-test/features/steps/users.py
```diff
@@ -4,7 +4,18 @@
...
+    base_orb_url = f"http://{configs.get('orb_address')}"
+else:
+    base_orb_url = f"https://{configs.get('orb_address')}"
+
+
+@given("the Orb user has a registered account")
+def register_orb_account(context):
+    email = configs.get('email')
+    password = configs.get('password')
+    register_account(email, password, 201)
```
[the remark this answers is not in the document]
manrodrigues: yes. I agree. Actually, on second thought, I think it makes more sense to create a test suite to test authentication/authorization and here assume that the account is already registered. I will remove this step and deploy this in the future.

python-test/features/steps/control_agents_ui.py
```diff
@@ -0,0 +1,121 @@
...
+import subprocess
+import shlex
+import re
+
+
+configs = TestConfig.configs()
+
+
+@given("the Orb user logs in through the UI")
+def logs_in_orb_ui(context):
```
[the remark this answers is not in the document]
manrodrigues: I moved to login_ui

python-test/features/steps/login_ui.py
manrodrigues: ok

python-test/features/steps/login_ui.py
manrodrigues: I liked. Good idea. Thanks

python-test/features/steps/login_ui.py
manrodrigues: OK. for now, I have separated the methods. I'll take a look at other possibilities, if you have any suggestions, please let me know

python-test/features/steps/users.py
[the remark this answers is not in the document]
stefano-daitan: There are many ways to make a failure scenario, which includes sending incorrect HTTP headers, malformed payload, etc.
This method will only handles failure status codes, so we may need to create new functions for the other scenarios.
Anyways, we can make the `request_status` an optional parameter with default value of 201, so it encapsulates knowledge about the happy path but still supports failure scenarios related to status code.
Another thing, let's rename that parameter to `expected_status_code`.

python-test/features/steps/login_ui.py
manrodrigues: ok

python-test/features/steps/control_agents_ui.py
```diff
@@ -0,0 +1,121 @@
...
+    WebDriverWait(context.driver, 3).until(
+        EC.presence_of_all_elements_located((By.XPATH, f"//div[contains(@class, 'agent-name') and contains(text(),"
+                                                       f"{context.agent_name})]")))[0].click()
+    agent_view_status = WebDriverWait(context.driver, 3).until(
+        EC.presence_of_all_elements_located(
+            (By.XPATH, "//label[contains(text(), 'Health Status')]/following::p")))[0].text
+    assert_that(agent_view_status, equal_to(status))
+
+
+def send_terminal_commands(command, separator=None, cwd_run=None):
```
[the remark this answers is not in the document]
manrodrigues: ok

python-test/features/steps/users.py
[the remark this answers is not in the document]
manrodrigues: I agree. I inserted the default value as suggested and also changed the name of the parameter.

python-test/features/steps/local_agent.py
```diff
@@ -20,6 +24,10 @@ def run_local_agent_container(context):
                 "ORB_CLOUD_MQTT_CHANNEL_ID": context.agent['channel_id'],
                 "ORB_CLOUD_MQTT_KEY": context.agent['key'],
                 "PKTVISOR_PCAP_IFACE_DEFAULT": interface}
+    if "localhost" in base_orb_url:
+        env_vars["ORB_CLOUD_ADDRESS"] = "localhost"
```
stefano-daitan, line 28: Isn't `base_orb_url` exactly equal to `localhost` in this case?

python-test/features/steps/ui_utils.py
```diff
@@ -0,0 +1,42 @@
+from hamcrest import *
+from selenium import webdriver
+
+
+def go_to_page(page, context, ignore_ssl_and_certificate_errors=True):
```
stefano-daitan, line 5: I understand that `ignore_ssl_and_certificate_errors` is a parameter, however by default we're ignoring possible SSL errors. But we should always verify the SSL certificate, and only in the specific case of using self-signed certificates we should ignore it.
This means we should check SSL certificates by default and only ignore errors if we're testing localhost.
If you'd like we can discuss the best approach to this.

python-test/features/steps/test_config.py
```diff
@@ -31,3 +31,15 @@ def _read_configs():
...
+def bypass_ssl_certificate_check():
+    ignore_ssl_and_certificate_errors = TestConfig.configs().get('ignore_ssl_and_certificate_errors', False)
+    if ignore_ssl_and_certificate_errors:
+        orb_url = f"http://{TestConfig.configs().get('orb_address')}"
+    else:
+        orb_url = f"https://{TestConfig.configs().get('orb_address')}"
+    return ignore_ssl_and_certificate_errors, orb_url
+
+
+bypass_ssl_certificate_check, base_orb_url = bypass_ssl_certificate_check()
```
stefano-daitan, line 45: Improve this as we talked.

python-test/features/steps/ui_utils.py
manrodrigues: Thanks for the tips. I think it's better this way.

python-test/README.md
```diff
@@ -25,7 +25,11 @@ Then fill in the correct values:
   - Orb user's password
 - **orb_address**:
   - Mandatory!
-  - URL of the Orb deployment. Do NOT include the protocol (`https://` or `mqtt://`).
+  - URL of the Orb deployment. Do NOT include the protocol (`https://`, `http://` or `mqtt://`).
+- **ignore_ssl_and_certificate_errors**:
+  - Bool
+  - Bypass SSL/certificate check
```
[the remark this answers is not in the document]
manrodrigues: Great!. I liked.

python-test/features/steps/ui_utils.py
```diff
@@ -0,0 +1,42 @@
...
+    """
+    options = webdriver.ChromeOptions()
+    if ignore_ssl_and_certificate_errors is True:
+        options.add_argument('--ignore-ssl-errors=yes')
+        options.add_argument('--ignore-certificate-errors')
+    options.add_argument("--start-maximized")
+    context.driver = webdriver.Chrome(options=options)
+    context.driver.get(str(page))
+    assert_that(context.driver.current_url, equal_to(f"{page}/auth/login"), "user not enabled to access orb "
+                                                                            "login page")
```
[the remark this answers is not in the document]
manrodrigues: Good point. Thanks!

python-test/features/steps/test_config.py
manrodrigues: Done

python-test/features/steps/users.py
```diff
@@ -32,3 +43,20 @@ def authenticate(user_email, user_password):
     assert_that(response.json(), has_key('token'), 'Authentication token not found in login response!')
 
     return response.json()['token']
+
+
+def register_account(user_email, user_password, expected_status_code=201):
+    """
+    Logs in to orb with given credentials
```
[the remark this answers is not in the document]
manrodrigues: ok

stefano-daitan (approved):
Thank you!

Issue #525, opened 2021-11-16
Create new landing page to create a new password


iatdaitan:

gpazuch:
@iatdaitan 
Is this the change password page the user should be redirected to after receiving the link for new password, after `Forgot Password` flow?

manrodrigues:
request password button is not working. Note that no request is sent

![reset-password-button](https://user-images.githubusercontent.com/78241475/146610119-a1e728fc-6654-420b-9d9b-25bcc4b15db1.gif)

manrodrigues:
once request new password is not working (as we discussed... maybe related with https://github.com/ns1labs/orb.live/issues/3), i'm not able to test this new page

manrodrigues:
We need to have a page to reset the password since now the email is working:

![page_reset](https://user-images.githubusercontent.com/78241475/171902585-f245e8ca-3255-4888-8848-412229308d99.gif)

PR #671, opened 2021-12-22
fix(orb-ui): #525 forgot password page

gpazuch:
fix(orb-ui): override nbResetPassword resetPassword method
fix(orb-ui): fix button outside form scope not submitting forgotPassword


iatdaitan (approved):
Approved. Good work.

@gpazuch is it possible to add a screenshot or even a gif of the final result as a comment?

PR #683, opened 2021-12-27
679 second policy

manrodrigues:


python-test/features/agentsProvider.feature
```diff
@@ -7,3 +7,17 @@ Feature: agent provider
             And the agent container is started
         Then the agent status in Orb should be online
             And the container logs should contain the message "sending capabilities" within 10 seconds
+
+
+    Scenario: Apply two policies to an agent
+        Given the Orb user logs in
+            And that an agent already exists and is online
```
leoparente, line 14: i think it should be "and it is online"

python-test/features/agentsProvider.feature
```diff
@@ -7,3 +7,17 @@ Feature: agent provider
...
+            And referred agent is subscribed to a group
+            And that a sink already exists
+            And that a policy already exists
+            And that a dataset using referred group, sink and policy ID already exists
+            And that container logs contains messages "scraped metrics for policy" and "managing agent policy from core" for applied policy within 180 seconds
+        When a new policy is created
+            And a new dataset is created using referred group, sink and policy ID
+        Then the container logs should contain messages "scraped metrics for policy" and "managing agent policy from core" for applied policy within 180 seconds
```
dscabral, line 22: I guees you should check if the new policy applied it's scrapping metrics... you can check the policy id / name on the logs... It should be present the two policies ids / names (e.g. 2021-12-27T11:20:20.118-0300	INFO	pktvisor/pktvisor.go:323	scraped metrics for policy	{"policy": "policy", "policy_id": "939a3af0-50d5-4338-ad43-2f1e9e68f7ca", "payload_size_b": 1829}
2021-12-27T11:20:20.118-0300	INFO	pktvisor/pktvisor.go:323	scraped metrics for policy	{"policy": "policy2", "policy_id": "10c3f357-a3bb-41f5-a829-d8358345df06", "payload_size_b": 1838})

python-test/features/steps/local_agent.py
```diff
@@ -101,6 +133,27 @@ def check_logs_contain_message(logs, expected_message):
...
+    :returns: (bool) whether expected message was found in the logs for expected policy
+    """
+
+    for log_line in logs:
+        log_line = safe_load_json(log_line)
+        if log_line is not None and log_line['msg'] == expected_message:
+            if 'id' in log_line.keys() and log_line['id'] == policy['id']:
+                return True
+            if 'policy_id' in log_line.keys() and log_line['policy_id'] == policy['id']:
+                return True
```
stefano-daitan, line 152: Why do we have `id` and `policy_id` as possible valid outcomes?

python-test/features/agentsProvider.feature
stefano-daitan: That's not necessary.

python-test/features/steps/local_agent.py
```diff
@@ -47,6 +54,25 @@ def check_agent_log(context, text_to_match, time_to_wait):
...
+      'time_to_wait} seconds')
+def check_agent_log_for_policy(context, text_to_match, time_to_wait):
+    time_waiting = 0
+    sleep_time = 0.5
+    timeout = int(time_to_wait)
+    text_found = False
+
+    while time_waiting < timeout:
+        logs = get_orb_agent_logs(context.container_id)
+        text_found = check_logs_contain_message_for_policy(logs, text_to_match, context.policy)
```
stefano-daitan, line 67: we should rename `context.policy` to `context.policy_id`

python-test/features/steps/local_agent.py
stefano-daitan: And do you think we should add policy-related knowledge to this `local_agent` module?

python-test/features/agentsProvider.feature
stefano-daitan: why 2 different policy IDs?
I see that `check_logs_contain_message_for_policy()` checks the logs like you mentioned.

weyrick: dscabral is correct - the goal of this test is to make sure that a second policy is applied and scraped correctly. unless we check for each policy separately, this test is not functioning correctly. but, it does appear that it tries to check id per policy.

python-test/features/steps/local_agent.py
weyrick: ok the log lines look like this
```
{"level":"info","ts":1640621757.9308138,"caller":"agent/comms.go:142","msg":"completed RPC subscription to group","name":"home","topic":"channels/65e72370-a778-48aa-8d2c-51e13412bad0/messages/fromcore"}
{"level":"info","ts":1640621757.94167,"caller":"policyMgr/manager.go:50","msg":"managing agent policy from core","action":"manage","name":"roku","dataset":"42d8e6dd-ac1a-47e6-b32b-a18746bddb42","backend":"pktvisor","id":"811c5953-9ab3-45be-8353-08136144944a","version":0}
{"level":"info","ts":1640621758.0065007,"caller":"policyMgr/manager.go:146","msg":"policy applied successfully","policy_id":"811c5953-9ab3-45be-8353-08136144944a","policy_name":"roku"}
{"level":"info","ts":1640621758.0065813,"caller":"policyMgr/manager.go:50","msg":"managing agent policy from core","action":"manage","name":"general","dataset":"aec80952-1434-48a0-8d9c-d91c0d9951ce","backend":"pktvisor","id":"0c4818ae-75c8-4359-aaf9-bdd8432cf577","version":0}
{"level":"info","ts":1640621758.0665298,"caller":"policyMgr/manager.go:146","msg":"policy applied successfully","policy_id":"0c4818ae-75c8-4359-aaf9-bdd8432cf577","policy_name":"general"}
```
So I see "managing agent policy from core" logs "id" instead of "policy_id" but this is a bug because we should be staying consistent here. Let's get a ticket open to fix the agent log message to use "policy_id".

However, I think it may also make more sense to look for the "policy applied successfully" message instead of "managing agent policy from core", which does use "policy_id".

weyrick (changes requested):

python-test/features/agentsProvider.feature
```diff
@@ -16,8 +16,14 @@ Feature: agent provider
...
+            And the container logs contains the message "policy applied successfully" referred to all applied policies within 10 seconds
+            And the container logs contains the message "managing agent policy from core" referred to all applied policies within 10 seconds
+            And the container logs contains the message "scraped metrics for policy" referred to all applied policies within 180 seconds
         When a new policy is created
             And a new dataset is created using referred group, sink and policy ID
-        Then the container logs should contain messages "scraped metrics for policy", "policy applied successfully" and "managing agent policy from core" for applied policy within 180 seconds
+        Then agent have 2 policies applied to it
+            And the container logs contains the message "policy applied successfully" referred to all applied policies within 10 seconds
+            And the container logs contains the message "managing agent policy from core" referred to all applied policies within 10 seconds
+            And the container logs contains the message "scraped metrics for policy" referred to all applied policies within 180 seconds
```
stefano-daitan, line 28: What if we only assert for logs and heartbeat after the 2nd policy has been attached to the agent?

python-test/features/steps/control_plane_policies.py
```diff
@@ -42,6 +50,36 @@ def new_policy(context):
...
+        sleep_time = 0.5
+        timeout = int(time_to_wait)
+        text_found = False
+        while time_waiting < timeout:
+            logs = get_orb_agent_logs(context.container_id)
+            text_found = check_logs_contain_message_for_policy(logs, text_to_match, policy_id)
+            if text_found is True:
+                break
+            time.sleep(sleep_time)
+            time_waiting += sleep_time
```
stefano-daitan, line 77: This method doesn't ensure that **all** applied policies show up in the log. It asserts if at least one does.

python-test/features/steps/control_plane_policies.py
```diff
@@ -42,6 +50,36 @@ def new_policy(context):
...
+    assert_that(len(context.list_agent_policies), equal_to(int(amount_of_policies)), f'Amount of policies applied to '
+                                                                                     f'this agent failed with '
+                                                                                     f'{amount_of_policies} policies')
+
+
+@step('the container logs contains the message "{text_to_match}" referred to all applied policies within {'
+      'time_to_wait} seconds')
+def check_agent_log_for_policy(context, text_to_match, time_to_wait):
+    assert_that(sorted(context.list_agent_policies), equal_to(sorted(context.policies_created.keys())))
+    for policy_id in context.list_agent_policies:
```
stefano-daitan, line 66: We don't need to loop through the logs N times to search for all policy IDs.

python-test/features/steps/control_plane_policies.py
```diff
@@ -42,6 +50,36 @@ def new_policy(context):
...
+        time_waiting = 0
+        sleep_time = 0.5
+        timeout = int(time_to_wait)
+        text_found = False
+        while time_waiting < timeout:
+            logs = get_orb_agent_logs(context.container_id)
```
stefano-daitan, line 72: We must get logs with timestamp later than when the 2nd policy was attached to the agent.

python-test/features/agentsProvider.feature
manrodrigues: tests now verify that all policies applied to the agent are actually applied and that they all publish metrics after the last policy was applied

python-test/features/steps/local_agent.py
manrodrigues: ok

python-test/README.md
```diff
@@ -12,6 +12,24 @@ Activate your virtual environment: `source name_of_virtualenv/bin/activate`
...
+
+Install ChromeDriver
+```
+$ wget https://chromedriver.storage.googleapis.com/$(google-chrome --product-version)/chromedriver_linux64.zip
+$ unzip chromedriver_linux64.zip
+$ sudo mv chromedriver /usr/bin/chromedriver
+$ sudo chown root:root /usr/bin/chromedriver
+$ sudo chmod +x /usr/bin/chromedriver
+```
+
```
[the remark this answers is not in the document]
manrodrigues: ok

python-test/features/steps/control_plane_policies.py
```diff
@@ -42,6 +50,36 @@ def new_policy(context):
     check_policies(context)
 
 
+@step('agent have {amount_of_policies} policies applied to it')
+def list_policies_applied_to_an_agent(context, amount_of_policies):
+    agent = get_agent(context.token, context.agent['id'])
+    context.list_agent_policies = list(agent['last_hb_data']['policy_state'].keys())
```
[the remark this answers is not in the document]
manrodrigues: this list only contain policies IDs, I already rename it

python-test/features/agentsProvider.feature
```diff
@@ -16,8 +16,14 @@ Feature: agent provider
             And that a sink already exists
             And that a policy already exists
             And that a dataset using referred group, sink and policy ID already exists
-            And that container logs contains messages "scraped metrics for policy", "policy applied successfully" and "managing agent policy from core" for applied policy within 180 seconds
+            And agent have 1 policies applied to it
```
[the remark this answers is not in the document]
manrodrigues: outdated

python-test/features/agentsProvider.feature
manrodrigues: good idea.

python-test/features/steps/control_plane_policies.py
manrodrigues: ok

python-test/features/steps/control_plane_policies.py
manrodrigues: yes. now we are using timestamp for matching logs.

python-test/features/steps/control_plane_policies.py
```diff
@@ -42,6 +50,52 @@ def new_policy(context):
     check_policies(context)
 
 
+@step('the container logs contains the message "{text_to_match}" referred to all applied policies within {'
+      'time_to_wait} seconds after applying all of them')
```
stefano-daitan, line 54: Looks like we're talking about 180 seconds after the policies have been applied, not logs messages printed after policies have been applied.

python-test/features/steps/control_plane_policies.py
```diff
@@ -42,6 +50,52 @@ def new_policy(context):
...
+
+    assert_that(text_found, is_(True),
+                f"Message '{text_to_match}' for policy "
+                f"'{set(context.list_agent_policies_id).difference(policies_with_message)}'"
+                f" was not found in the agent logs!")
+
+
+@step('the container logs contains the message "{text_to_match}" referred to all applied policies within {'
+      'time_to_wait} seconds')
+def check_agent_log_for_policy(context, text_to_match, time_to_wait):
```
stefano-daitan, line 78: This function could simply call the function above.

python-test/features/steps/control_plane_policies.py
```diff
@@ -137,3 +191,28 @@ def delete_policy(token, policy_id):
...
+    :returns: (bool) whether expected message was found in the logs for expected policy
+
+
+    """
+    policies_with_message = set()
+    for log_line in logs:
+        log_line = safe_load_json(log_line)
+        if log_line is not None and log_line['msg'] == expected_message:
+            for policy_id in policies_list:
+                if 'policy_id' in log_line.keys() and log_line['policy_id'] == policy_id and log_line['ts'] > timestamp_considered:
```
stefano-daitan, line 214: These `if` statements are fuzzy and made the code confusing.
Let's create a helper function `isExpectedLogLine()` in this module that tests log lines.
Log lines should:
- not be None
- have a 'msg' property that matches the `expected_message` string.
- Have a 'ts' property whose value is greater than `timestamp_considered`
- have a property 'policy_id' that is also contained in the `policies_list` list.

python-test/features/steps/control_plane_sink.py
```diff
@@ -36,10 +37,20 @@ def create_sink(context):
...
+    timeout = int(time_to_wait)
     sink_id = context.sink["id"]
-    get_sink_response = get_sink(context.token, sink_id)
+
+    while time_waiting < timeout:
+        get_sink_response = get_sink(context.token, sink_id)
+        if get_sink_response['state'] == status:
+            break
+        time.sleep(sleep_time)
+        time_waiting += sleep_time
```
stefano-daitan, line 52: Many functions use this `within X seconds` pattern. In future PRs we should start thinking about a function that encapsulates this retry logic.

manrodrigues: Sure. Good ideia.

python-test/features/agentsProvider.feature
```diff
@@ -7,3 +7,18 @@ Feature: agent provider
...
+            And that an agent already exists and is online
+            And referred agent is subscribed to a group
+            And that a sink already exists
+            And that a policy already exists
+            And that a dataset using referred group, sink and policy ID already exists
+        When a new policy is created
+            And a new dataset is created using referred group, sink and policy ID
+        Then this agent's heartbeat shows that all 2 policies have been successfully applied
+            And the container logs contains the message "policy applied successfully" referred to each policy within 10 seconds
+            And the container logs that were output after all policies have been applied contains the message "scraped metrics for policy" referred to each policy within 180 seconds
```
stefano-daitan, line 23: typo: contains -> contain

python-test/features/steps/control_plane_policies.py
```diff
@@ -42,6 +50,19 @@ def new_policy(context):
...
+@step('the container logs that were output after all policies have been applied contains the message "{'
+      'text_to_match}" referred to each policy within {time_to_wait} seconds')
+def check_agent_logs_for_policies(context, text_to_match, time_to_wait):
+    check_agent_log_for_policies(text_to_match, time_to_wait, context.container_id, context.list_agent_policies_id,
+                                 context.dataset_applied_timestamp)
+
+
+@step('the container logs contains the message "{text_to_match}" referred to each policy within {'
+      'time_to_wait} seconds')
+def check_agent_logs_for_policies(context, text_to_match, time_to_wait):
```
stefano-daitan, line 62: There are 2 functions with the same signature (same name `check_agent_logs_for_policies` and same input parameters). Isn't this going to cause an issue when python executes the code?

python-test/features/agentsProvider.feature
manrodrigues: thanks

python-test/features/steps/control_plane_policies.py
manrodrigues: Since the function does the same thing in both places, behave does not generate redundancy if the more specific function (with extra parameters) is defined first, but in any case, I agree that it is better to keep the pattern. We will keep different names for all functions.

python-test/features/steps/control_plane_policies.py
```diff
@@ -192,27 +159,79 @@ def delete_policy(token, policy_id):
...
+    as soon as a log is found with the expected message for each applied policy.
 
     :param (list) logs: list of log lines
     :param (str) expected_message: message that we expect to find in the logs
-    :param (str) policies_list: list with all policy id applied to the agent
-    :param timestamp_considered: timestamp from which the log will be considered.
-                                                                Default: timestamp at which behave execution is started
-    :returns: (bool) whether expected message was found in the logs for expected policy
+    :param (list) list_agent_policies_id: list with all policy id applied to the agent
+    :param (float) considered_timestamp: timestamp from which the log will be considered
```
stefano-daitan, line 170: Sorry to only catch this now, but timestamp should be an integer, no?

stefano-daitan (approved):
Good job, just a minor comment about docs.

python-test/features/steps/control_plane_policies.py
manrodrigues: No. Timestamp is a float. 
https://docs.python.org/3/library/datetime.html

` datetime.timestamp()
Return POSIX timestamp corresponding to the datetime instance. The return value is a float similar to that returned by time.time().`

Issue #641, opened 2021-12-17
Change the color of MOCK text on provisioning command


iatdaitan:

manrodrigues:
Tested using last commit - 0fe8fef3201cc31e7a7da8faed1aecb17a0185e0

<br>

![image](https://user-images.githubusercontent.com/78241475/148785665-6398e26e-c43d-4e45-a7dd-00da3cd738f6.png)

<br>

![image](https://user-images.githubusercontent.com/78241475/148785718-7f500f2d-81f0-47a1-9107-dc21a74c141f.png)

PR #715, opened 2022-01-03
feat(orb-ui): #641 change color of 'mock' string

gpazuch:
image -- using \<mark\/\> tag

mark text where user could/should change interface


gpazuch:
Alternatively, we could change just the font to a brighter color instead of <mark>ing it.

weyrick:
I think it's good enough to change the text color, not have the background changed. Also the text "Replace mock interface with host network interface" should be that same text color probably?

gpazuch:
image
Oh ok, lemme take a screenshot of just the text color.

gpazuch:
@weyrick 
image

gpazuch:
image
@weyrick

gpazuch:
image

weyrick (approved):
awesome

PR #726, opened 2022-01-06
int_tests - automate account registration with credentials required (email and password) if they are not yet registered

manrodrigues:


leoparente (dismissed):

iatdaitan (approved):
Everything seems to be really straightforward. Good job.

PR #729, opened 2022-01-06
fix(orb-ui): #692 Dataset View

### LINKED ISSUE #692: Datasets: Its not easy to deploy a dataset connection
Datasets details displays linked group, policy and sink by ID, so its not clear what element should be changed or removed (note that even when user have few policies, groups or sinks, the ID is not displayed for any of them).  Also the user is not able to sort by ID, so this shouldn't be the identifier.

<br>

image

<br>

image

<br>

image

<br>

expected result:
group, sink and policy related to a datasets should be displayed by their own name

gpazuch:
Resolves #692

### DIFF
--- a/ui/src/app/common/services/agents/agent.groups.service.ts
+++ b/ui/src/app/common/services/agents/agent.groups.service.ts
@@ -85,7 +85,7 @@ export class AgentGroupsService {
       );
   }
 
-  getAgentGroupById(id: string): any {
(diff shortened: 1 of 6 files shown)


gpazuch:
image
@iatdaitan while we're at it, should I remove the red asterisks in this view page?
I mean, it does make a difference on edit and add, but here as well?

PR #732, opened 2022-01-07
Feat/orb policy associate schema version

dscabral:
Create schema version association between backend and policy on policy creation
Update policies openapi spec
Update fleet list backends endpoint to return schema version
Update fleet openapi spec
Create a new table column on postgres init file


fleet/api/http/openapi.yaml
```diff
@@ -770,6 +770,10 @@ components:
             type: string
             description: Description of this backend
             example: pktvisor observability agent from pktvisor.dev
+          schema_version:
+            type: string
+            description: Backend version
```
weyrick, line 775: "Version of the schema for this Backend"

fleet/backend/pktvisor/types.go
```diff
@@ -0,0 +1,3 @@
+package pktvisor
+
+const CurrentSchemaVersion = "1.0"
```
weyrick, line 3: You can also use this in `agent/backend/pktvisor/pktvisor.go` line 155

policies/api/http/responses.go
```diff
@@ -16,6 +16,7 @@ type policyRes struct {
 	Description string         `json:"description,omitempty"`
 	Tags        types.Tags     `json:"tags"`
 	Backend     string         `json:"backend"`
+	Schema      string         `json:"schema"`
```
weyrick, line 19: This should be SchemaVersion and `schema_version` too, for consistency

policies/postgres/init.go
```diff
@@ -51,6 +51,7 @@ func migrateDB(db *sqlx.DB) error {
 						orb_tags       JSONB NOT NULL DEFAULT '{}',
 
 						backend        TEXT NOT NULL,
+                        schema_version TEXT NOT NULL,
```
weyrick, line 54: So for the first time we cannot just update "policies_1", we have to make a migration here with "policies_2". See this for an example https://github.com/mainflux/mainflux/blob/aa014c21915b42764e070276a1e05c49a43feac7/things/postgres/init.go#L83

policies/api/http/requests.go
```diff
@@ -24,14 +24,15 @@ const (
...
+	SchemaVersion string         `json:"schema_version"`
+	Policy        types.Metadata `json:"policy,omitempty"`
+	Tags          types.Tags     `json:"tags"`
+	Format        string         `json:"format,omitempty"`
+	PolicyData    string         `json:"policy_data,omitempty"`
+	Description   string         `json:"description"`
+	token         string
 }
 
 func (req addPolicyReq) validate() error {
```
weyrick, line 38: In this function, if SchemaVersion is empty we should default it to "1.0"

weyrick (changes requested):

weyrick (approved):

Issue #737, opened 2022-01-10
Not able to register new account ("REGISTER" button is not working)


manrodrigues:
Button is not sending any request to endpoint users (version -> 0fe8fef3201cc31e7a7da8faed1aecb17a0185e0)

gpazuch:
![clickwrap](https://user-images.githubusercontent.com/1490938/148773785-3a68d63f-3916-411b-8d68-cbc3aa21585e.gif)

manrodrigues:
@gpazuch  Two comments about things that I noticed.
1) one that I can see in your gif as well, that is, even when user can register an account, a token error is displayed. If token is alright, this error must not appear, but if there is a problem with token, user must not be enabled to register an account.
<br>

![image](https://user-images.githubusercontent.com/78241475/148985193-1e9b302f-347d-43b0-b484-034371a1b11a.png)

<br>


![register_account2](https://user-images.githubusercontent.com/78241475/148985901-59e8973a-eb36-4280-a418-92dbb80a774a.gif)

2)if user forget to click on acceptance term, help message is correctly displayed, but user is not enable do register account anymore, because register button keeps disabled

<br>

![register_account](https://user-images.githubusercontent.com/78241475/148986148-1464d1b5-6b2a-4f85-8d10-8098939500e4.gif)

gpazuch:
![image](https://user-images.githubusercontent.com/1490938/149349527-5a03c19e-54e8-4c85-85c6-fcdfa27575f2.png)

It seems I'm always getting an error response on register endpoint, it seems to expect a token from the register endpoint as well as the authenticate endpoint.

I'm looking for a fix

PR #740, opened 2022-01-10
fix(orb-ui): #737 register button not working

gpazuch:
Fix for pactsafe on --prod

@manrodrigues try and set `
PS_GROUP_KEY=group-rkjy8a9uy
` on ui/.env file


gpazuch:
This does not hide contract box on development build

iatdaitan (approved):

PR #750, opened 2022-01-12
refactor(orb-agent): enhancement on log messages

dscabral:
Remove retry on pktvisor request
Create more wrappers for errors messages related to mqtt errors
Show orb-agent version on start up

### DIFF
--- a/agent/comms.go
+++ b/agent/comms.go
@@ -81,7 +81,7 @@ func (a *orbAgent) startComms(config config.MQTTConfig) error {
 	var err error
 	a.client, err = a.connect(config)
 	if err != nil {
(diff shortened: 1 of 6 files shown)


agent/comms.go
```diff
@@ -81,8 +81,8 @@ func (a *orbAgent) startComms(config config.MQTTConfig) error {
 	var err error
 	a.client, err = a.connect(config)
 	if err != nil {
-		a.logger.Error("connection failed", zap.Error(err))
-		return err
+		a.logger.Error("connection failed", zap.String("key", config.Key), zap.String("channel", config.ChannelID), zap.String("id", config.Id), zap.Error(err))
```
weyrick, line 84: we can't log the key, it's essentially a password. and let's use "agent_id" and never just "id" because it's hard to search for that later, or understand which id it is

weyrick (changes requested):

weyrick (approved):

Issue #368, opened 2021-10-08
Policy management: all required fields should have "*"


manrodrigues:
Handler Label is without "*"

weyrick:
this is probably outdated?

gpazuch:
@iatdaitan I think this shouldn't take longer than 1.5~2 hours for all pages we currently have

manrodrigues:
![image](https://user-images.githubusercontent.com/78241475/151806787-d4cbc435-4c0f-40a8-8932-79a95dd9a795.png)

![image](https://user-images.githubusercontent.com/78241475/151806822-9bf05710-dbf2-4637-a4a7-b61697e070e7.png)

![image](https://user-images.githubusercontent.com/78241475/151806863-03101f95-5121-4963-9ece-9a6e1148e677.png)

Last commit: https://github.com/ns1labs/orb/commit/fa3629fb4dd4adc821d788fade84c2ef6c3c81fe

PR #753, opened 2022-01-12
fix(orb-ui): #368 all required fields should have asterisk

gpazuch:
fix(orb-ui): add missing asterisk to handler label field


gpazuch:
@manrodrigues this should be a simple one liner, it seems only handler label was missing the asterisk.

gpazuch:
@manrodrigues this should be a simple one liner, it seems only handler label was missing the asterisk.

Just lemme fix the ancestry of this branch

manrodrigues (approved):

PR #758, opened 2022-01-13
fix(orb-ui): fix error message during register

gpazuch:
* authentication module no longer expects valid token on /users endpoint

* remove code for redirection on /users endpoint response - force auth


iatdaitan (approved):
LGTM

manrodrigues:
@gpazuch 
i'm able to register without accept the contract now. My localhost is not pointing to any contract, although I have the variables registered

<br>

image

<br>

register_without_contract

PR #770, opened 2022-01-17
fix(orb-ui): expose pactsafe env vars to docker space in local builds

gpazuch:
* building orb-ui via `make ui` -> make sure PS_SID and PS_GROUP_KEY are present in the environment.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -112,9 +112,12 @@ agent:
 
 ui:
 	cd ui/ && docker build \
-    		  --tag=$(DOCKERHUB_REPO)/$(DOCKER_IMAGE_NAME_PREFIX)-ui:$(REF_TAG) \
-    		  --tag=$(DOCKERHUB_REPO)/$(DOCKER_IMAGE_NAME_PREFIX)-ui:$(ORB_VERSION) \
-    		  --tag=$(DOCKERHUB_REPO)/$(DOCKER_IMAGE_NAME_PREFIX)-ui:$(ORB_VERSION)-$(COMMIT_HASH) \
-    		  -f docker/Dockerfile .
+		--build-arg ENV_PS_SID=${PS_SID} \
+		--build-arg	ENV_PS_GROUP_KEY=${PS_GROUP_KEY} \
+		--build-arg ENV=${ENVIRONMENT} \
+		--tag=$(DOCKERHUB_REPO)/$(DOCKER_IMAGE_NAME_PREFIX)-ui:$(REF_TAG) \
+		--tag=$(DOCKERHUB_REPO)/$(DOCKER_IMAGE_NAME_PREFIX)-ui:$(ORB_VERSION) \
+		--tag=$(DOCKERHUB_REPO)/$(DOCKER_IMAGE_NAME_PREFIX)-ui:$(ORB_VERSION)-$(COMMIT_HASH) \
+		-f docker/Dockerfile .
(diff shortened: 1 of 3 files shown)


gpazuch:
@manrodrigues this should generate ui docker image w/ pactsafe, instead of serving ui locally via node and setting ui/.env

gpazuch:
@manrodrigues this should generate ui docker image w/ pactsafe, instead of serving ui locally via node and setting ui/.env.

make sure to run

`export PS_SID=${your_value_here}`
and 
`export PS_GROUP_KEY=${your_value_here}`

prior to `make ui`

Makefile
```diff
@@ -112,9 +112,12 @@ agent:
 
 ui:
 	cd ui/ && docker build \
-    		  --tag=$(DOCKERHUB_REPO)/$(DOCKER_IMAGE_NAME_PREFIX)-ui:$(REF_TAG) \
-    		  --tag=$(DOCKERHUB_REPO)/$(DOCKER_IMAGE_NAME_PREFIX)-ui:$(ORB_VERSION) \
-    		  --tag=$(DOCKERHUB_REPO)/$(DOCKER_IMAGE_NAME_PREFIX)-ui:$(ORB_VERSION)-$(COMMIT_HASH) \
-    		  -f docker/Dockerfile .
+		--build-arg ENV_PS_SID=${PS_SID} \
+		--build-arg	ENV_PS_GROUP_KEY=${PS_GROUP_KEY} \
```
weyrick, line 116: please fix this white space

weyrick (changes requested):

docker/docker-compose.yml
```diff
@@ -461,6 +461,9 @@ services:
     environment:
       MF_UI_PORT: ${MF_UI_PORT}
       MF_UI_MQTT_WS_URL: ${MF_UI_MQTT_WS_URL}
+      PS_SID: ${PS_SID}
+      PS_GROUP_KEY: ${PS_GROUP_KEY}
+
```
dscabral, line 466: Could you put values on `.env` file for those env variables??

dscabral (changes requested):

docker/docker-compose.yml
gpazuch: I shouldn't commit the values for those. Each user that downloads and spins their own Orb copy, can, in fact, sign up for IronClad and then add their own keys.

Staging environment contains the keys for testing, while live environment should be built with definitive keys for this specific environment.

Hence, the values are commited in orb.live's repo, but not available in orb's repo.

The correct procedure, either via .env file(in case UI is able to be built via docker-compose), or environment vars, is explained above.

I suspect settings those to .env won't take effect, since we need those values at the moment the image is built - and orb-ui is compiled.

dscabral (approved):

gpazuch:
@weyrick there is one known issue in local builds/hosting of orb

* when loading register page, rendering of contract name/link next to the checkbox is not always present.

I suspect that the component for PS is not ready at the time register page finishes loading, causing this issue.

The image attached shows this behavior.

image

Makefile
gpazuch: @weyrick I think this request must be resolved by you prior to merging.
I left a comment in the PR as well on current behavior.

weyrick (approved):
lgtm

PR #774, opened 2022-01-18
Fix policies migrations error

dscabral:
Set valid value for postgres 'schema_version' column on script for migration


policies/postgres/init.go
```diff
@@ -86,7 +86,7 @@ func migrateDB(db *sqlx.DB) error {
 				Id: "policies_2",
 				Up: []string{
 					`ALTER TABLE IF EXISTS agent_policies ADD COLUMN IF NOT EXISTS
-					 schema_version TEXT NOT NULL`,
+					 schema_version TEXT NOT NULL DEFAULT ''`,
```
weyrick, line 89: well let's make default '1.0'

weyrick (changes requested):

weyrick (approved):

PR #781, opened 2022-01-20
Feat/orb agent handle multiple datasets

dscabral:


agent/policyMgr/manager.go
```diff
@@ -75,6 +77,17 @@ func (a *policyManager) ManagePolicy(payload fleet.AgentPolicyRPCPayload) {
 					a.logger.Warn("policy failed to ensure dataset id", zap.String("policy_id", payload.ID), zap.String("policy_name", payload.Name), zap.String("dataset_id", payload.DatasetID), zap.Error(err))
 				}
 			}
+			// if policy already exist and has no version upgrade, has no need to apply it again on pktvisor
```
weyrick, line 80: this code is backend agnostic, so let's not mention pktvisor

agent/policyMgr/manager.go
```diff
@@ -75,6 +77,17 @@ func (a *policyManager) ManagePolicy(payload fleet.AgentPolicyRPCPayload) {
...
+			currentPolicy, err := a.repo.Get(payload.ID)
+			if err != nil {
+				a.logger.Error("failed to retrieve policy", zap.String("policy_id", payload.ID), zap.Error(err))
```
weyrick, line 83: need a `return` after this, not just error log

agent/policyMgr/manager.go
```diff
@@ -75,6 +77,17 @@ func (a *policyManager) ManagePolicy(payload fleet.AgentPolicyRPCPayload) {
...
+			}
+			if currentPolicy.Version >= pd.Version {
+				a.logger.Info("policy already applied or with greater version, skipping", zap.String("policy_id", pd.ID), zap.String("policy_name", pd.Name), zap.String("attempted version", fmt.Sprint(pd.Version)), zap.String("current version", fmt.Sprint(currentPolicy.Version)))
```
weyrick, line 86: how about "we have a better version of this policy, skipping". the log keys should not have spaces, so "attempted_version" and "current_version"

agent/policyMgr/manager.go
```diff
@@ -91,6 +104,13 @@ func (a *policyManager) ManagePolicy(payload fleet.AgentPolicyRPCPayload) {
 		} else {
 			// attempt to apply the policy to the backend. status of policy application (running/failed) is maintained there.
 			be := backend.GetBackend(payload.Backend)
+			if updatePolicy {
```
weyrick, line 107: again, this code is backend diagnostic so we shouldn't mention pktvisor or force pktvisor semantics.

agent/policyMgr/manager.go
```diff
@@ -91,6 +104,13 @@ func (a *policyManager) ManagePolicy(payload fleet.AgentPolicyRPCPayload) {
...
+				// pktvisor has no update endpoint, so to update a policy it's necessary remove it and apply a new version
+				err := be.RemovePolicy(pd)
+				if err != nil {
+					a.logger.Warn("policy failed to remove", zap.String("policy_id", payload.ID), zap.String("policy_name", payload.Name), zap.Error(err))
+				}
+			}
 			a.applyPolicy(payload, be, &pd)
```
weyrick, line 114: what we should do instead is to include the `updatePolicy` bool as a parameter to `applyPolicy`. Then you can move the "remove first" logic to this pktvisor specific method implementation https://github.com/ns1labs/orb/blob/1a9ab8a563ba47c9a10f4546e259a4c44fe70e87/agent/backend/pktvisor/pktvisor.go#L149

agent/policyMgr/manager.go
```diff
@@ -114,7 +134,7 @@ func (a *policyManager) ManagePolicy(payload fleet.AgentPolicyRPCPayload) {
 		// Remove policy from orb-agent local repo
 		err = a.repo.Remove(pd.ID)
 		if err != nil {
-			a.logger.Warn("policy failed to remove local", zap.String("policy_id", pd.ID),  zap.String("policy_name", pd.Name), zap.Error(err))
+			a.logger.Warn("policy failed to remove local", zap.String("policy_id", pd.ID), zap.String("policy_name", pd.Name), zap.Error(err))
```
weyrick, line 137: is this just whitespace?

weyrick (changes requested):

PR #803, opened 2022-01-27
Tests/integration tests

manrodrigues:
- Apply the same policy twice to the agent
- Remove group (invalid dataset, agent logs)
- Check if email and password are required fields

### DIFF
--- a/python-test/features/steps/login.py
+++ b/python-test/features/steps/login.py
@@ -21,29 +21,25 @@ def check_registered_account(context):
 @when("user request account registration {email} email, {password} password, {username} user name and {company} "
       "company name")
 def check_account_input(context, email, password, username, company):
-    assert_that(email, any_of(equal_to('with'), equal_to('without')),
-                "Not expected option to email")
-    assert_that(password, any_of(equal_to('with'), equal_to('without')),
-                "Not expected option to password")
-    assert_that(username, any_of(equal_to('with'), equal_to('without')),
-                "Not expected option to username")
-    assert_that(company, any_of(equal_to('with'), equal_to('without')),
-                "Not expected option to company")
+    inputs = {'email': email, 'password': password, 'username': username, 'company': company}
+    for key, value in inputs.items():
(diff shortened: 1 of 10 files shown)


python-test/features/steps/login.py
```diff
@@ -18,7 +18,40 @@ def check_registered_account(context):
...
+        account_input['password'] = configs.get('password')
+    if username == "with":
+        account_input['username'] = f"test_user {random_string(3)}"
+    if company == "with":
+        account_input['company'] = f"test_company {random_string(3)}"
+    if email == "without" or password == "without":
+        account_input['reg_status'] = 400
+        account_input['auth_status'] = 400
+    if email == "with" and password == "without":
+        account_input['auth_status'] = 403
```
iatdaitan, line 46: What if we rewrite this part to something like that ?

`
kvargs = {
    'email': email, 
    'password': password, 
    'username': username, 
    'company': company
}

for key, value in kvargs.items():
   assert_that(value, any_of(equal_to('with'), equal_to('without')),
                f"Not expected option to { key }")
        
account_input = {
    'email': None, 
    'password': None, 
    'company': None, 
    'username': None, 
    'reg_status': 201,
    'auth_status': 201
}

if email == "with":
    account_input['email'] = f"test_email_{random_string(3)}@email.com"
    
    if password == "without":
        account_input['auth_status'] = 403
elif password == "without":
    account_input['reg_status'] = 400
    account_input['auth_status'] = 400
    
if password == "with":
    account_input['password'] = configs.get('password')
    
if username == "with":
    account_input['username'] = f"test_user {random_string(3)}"
    
if company == "with":
    account_input['company'] = f"test_company {random_string(3)}"
`

manrodrigues:
```
elif password == "without":
account_input['reg_status'] = 400
account_input['auth_status'] = 400
```

exactly that way it doesn't work because we have to consider when is without email and with password.. but I adapted your idea..
If you have another suggestion, please share

iatdaitan (approved):

PR #816, opened 2022-01-28
Feat/orb agent reset

dscabral:


fleet/api/http/metrics.go
```diff
@@ -22,6 +22,28 @@ type metricsMiddleware struct {
...
 
+func (m metricsMiddleware) ResetAgent(ct context.Context, token string, agentID string) error {
+	ownerID, err := m.identify(token)
+	if err!= nil {
+		return err
+	}
+
+	defer func(begin time.Time) {
+		labels := []string{
+			"method", "viewOwnerByChannelIDInternal",
```
weyrick, line 33: wrong method

fleet/api/http/transport.go
```diff
@@ -90,6 +90,11 @@ func MakeHandler(tracer opentracing.Tracer, svcName string, svc fleet.Service) h
 		decodeListBackends,
 		types.EncodeResponse,
 		opts...))
+	r.Get("/agents/:id/reset", kithttp.NewServer(
```
weyrick, line 93: let's make the endpoint /agents/:id/rpc/reset and make it a POST instead of a GET (because GET usually means "read only")

fleet/agent_service.go
```diff
@@ -62,6 +62,20 @@ func (svc fleetService) ViewAgentByID(ctx context.Context, token string, thingID
...
+	if err != nil {
+		return err
+	}
+
+	agent, err := svc.agentRepo.RetrieveByID(ctx, ownerID, agentID)
+	if err != nil {
+		return err
+	}
+
+	return svc.agentComms.NotifyAgentReset(agent.MFChannelID, true, "")
```
weyrick, line 76: let's make the default reason "Reset initiated from control plane" for now

agent/rpc_from.go
```diff
@@ -98,6 +98,12 @@ func (a *orbAgent) handleDatasetRemoval(rpc fleet.DatasetRemovedRPCPayload) {
 	a.removeDatasetFromPolicy(rpc.DatasetID, rpc.PolicyID)
 }
 
+func (a *orbAgent) handleAgentReset(payload fleet.AgentResetRPCPayload) {
+	// TODO implement logic for agent and backend reset
```
weyrick, line 102: we can remove the comments? and pass in the `payload` here so `Restart` has access to its contents. it can check FullReset there for now.

weyrick (changes requested):

weyrick (approved):

PR #96, opened 2021-08-10
ORB-69: Sink Add Page

gpazuch:
* inject router to sinks.add.component
* change <create sink> button to route to sinks.add.component
* add notifications for success|failure upon creation
* modal to page conversion
* use strings from STRINGS definition
* sinks.add.component able to support add and edit
* merge sink.details improvements pr #97 

closes #69
advance work in #51 by making sinks.add.component reusable
advance work in #65 removing unused css rules

PR #837, opened 2022-02-07
fix(orb-ui): #96 pactsafe parameters

gpazuch:
* require two names, space separated, for full user name
* send full user name as signer id when contract agreement
* pactsafe confirmation email set to true - user should receive a copy of contract


gpazuch:
@manrodrigues the pactsafe confirmation email and contract box should show only on production builds.

the name validator for two space separated names is in place for all builds, it should also verify for combined lenght of names.

iatdaitan (approved):

PR #841, opened 2022-02-07
agent subscription to multiple groups (orb-tags)

manrodrigues:
- I changed the logic from groups and agents creation in order to enable multiple and defined tags
- Scenarios are:
       - subscribe an agent to multiple groups created before agent provision
       - subscribe an agent to multiple groups created after agent provision
      - create agent without tags
      - create agent with multiple tags
      - create agent group without tags
      - create agent group with multiple tags

### DIFF
--- a/python-test/features/agentProviderUi.feature
+++ b/python-test/features/agentProviderUi.feature
@@ -5,14 +5,14 @@ Feature: Create agents using orb ui
         Given the Orb user logs in through the UI
             And that fleet Management is clickable on ORB Menu
             And that Agents is clickable on ORB Menu
-        When a new agent is created through the UI with defined orb tag(s): region:br, demo:true, ns1:true
+        When a new agent is created through the UI with region:br, demo:true, ns1:true orb tag(s)
         Then the agents list and the agents view should display agent's status as New within 10 seconds
 
     Scenario: Provision agent
         Given the Orb user logs in through the UI
             And that fleet Management is clickable on ORB Menu
             And that Agents is clickable on ORB Menu
-        When a new agent is created through the UI with random orb tag(s): 2 tag(s)
+        When a new agent is created through the UI with 2 orb tag(s)
             And the agent container is started using the command provided by the UI on port default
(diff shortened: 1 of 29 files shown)


python-test/features/agentProviderUi.feature
```diff
@@ -5,14 +5,14 @@ Feature: Create agents using orb ui
...
-        When a new agent is created through the UI
+        When a new agent is created through the UI with defined orb tag(s): region:br, demo:true, ns1:true
         Then the agents list and the agents view should display agent's status as New within 10 seconds
 
     Scenario: Provision agent
         Given the Orb user logs in through the UI
             And that fleet Management is clickable on ORB Menu
             And that Agents is clickable on ORB Menu
-        When a new agent is created through the UI
+        When a new agent is created through the UI with random orb tag(s): 2 tag(s)
```
leoparente, line 15: don't get the "2 tags()". what that does mean?

python-test/features/steps/page_objects.py
```diff
@@ -0,0 +1,19 @@
+#XPATHs
```
leoparente, line 1: this was a good idea. maybe add a namespace to the variables so we can know that it comes from this file.

like a prefix to know that the variables are coming from this file.  ex: orb_page_*

Or add a class and export variables through that. example OrbPage object and you get `OrbPage.fleet_management_left_menu` and so on.

I think that second approach is more Object Oriented.

leoparente (changes requested):

python-test/features/agentProviderUi.feature
manrodrigues: it means the amount of tags that must be added to the agent or to the group..

python-test/features/steps/page_objects.py
manrodrigues: oh, interesting and good idea.. i'm gonna working on that

leoparente (dismissed):

leoparente (approved):

PR #843, opened 2022-02-07
fix(orb-ui): agent policy fixes

gpazuch:
* make handler editing a separate modal
* display added handlers on agent policy page
* fix qname string/array issue


iatdaitan (approved):
LGTM

PR #846, opened 2022-02-07
fix(orb-ui): Add cancel button to all stepper controls in forms

gpazuch:
* Agent Create/Edit
* Agent Group/Edit
* Agent Policy Create/Edit
* Dataset Create/Edit
* Sink Create/Edit

### DIFF
--- a/ui/src/app/pages/datasets/add/dataset.add.component.html
+++ b/ui/src/app/pages/datasets/add/dataset.add.component.html
@@ -92,6 +92,14 @@
             <hr/>
             <div class="d-flex justify-content-end">
               <button
+                (click)="goBack()"
+                data-orb-qa-id="cancel"
+                ghost
+                nbButton
+                status="primary"
+                type="button">Cancel
+              </button>
+              <button
                 data-orb-qa-id="cancel"
                 ghost
                 nbButton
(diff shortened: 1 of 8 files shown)


ui/src/app/pages/datasets/add/dataset.add.component.html
```diff
@@ -91,6 +91,14 @@ <h4>{{isEdit ? 'Edit Dataset' : 'New Dataset'}}</h4>
             </nb-form-field>
             <hr/>
             <div class="d-flex justify-content-end">
+              <button
+                (click)="goBack()"
+                data-orb-qa-id="cancel"
+                ghost
+                nbButton
+                status="primary"
+                type="button">Cancel
```
iatdaitan, line 100: @gpazuch it would be good to follow the same approach of the agent group add component. May we should create an object to hold those label information? It will allow us to change all affected pages at once by only changing on file.

What do you think about it?

iatdaitan: The same applies to the other files of this PR.

gpazuch: Sure thing. There is a place for that.

iatdaitan: Would you address this fix on this PR, or should I open an issue to it?

gpazuch: I'm fixing it right away

gpazuch: Fixed

manrodrigues (approved):

PR #857, opened 2022-02-09
Bug/agent group subscription fail

dscabral:
Apply new policies, datasets, and groups on agents after editing agents or agent_groups on the fleet management (Only the channel was being subscribed before) 
When editing a tag from an agent, on channel connection, if mainflux returned a 409 error for one of the channels the new channels which really need to be applied were not happening. Now we're just skipping 409 errors from the channel connection.


fleet/agent_group_service.go
```diff
@@ -121,6 +123,28 @@ func (svc fleetService) EditAgentGroup(ctx context.Context, token string, group
 	// append both lists and remove duplicates
 	// need to unsubscribe the agents who are no longer matching with the group
 	list := removeDuplicates(listSub, listUnsub)
+
+	// connect all agents to the group channel (check the already connected and connect the new ones)
+	if !reflect.DeepEqual(listSub, listUnsub) {
+		for _, a := range listSub {
```
weyrick, line 129: this will connect the new ones but don't we also have to disconnect the `listUnsub`?

dscabral: I haven't tested the disconnect feature. I'm not sure if it'll cause an error on the agent... or if disconnecting it's a graceful thing.

agent/rpc_from.go
```diff
@@ -17,11 +17,25 @@ func (a *orbAgent) handleGroupMembership(rpc fleet.GroupMembershipRPCPayload) {
 	// if this is the full list, reset all group subscriptions and subscribed to this list
 	if rpc.FullList {
 		a.unsubscribeGroupChannels()
+		policies, err := a.policyManager.GetRepo().GetAll()
```
weyrick, line 20: Unfortunately I think this is too destructive - if I understand correctly, it will remove all running policies every time there is a group change. The problem with this is that the time series will now have a gap in the data, as it throws out the data it had collected for the last minute, and it will take another minute to get it back. What we should really be doing is getting the new list of policies, and then compare that list to the list that is currently running - then add new policies and remove only those no longer in our group. If you think that's too hard then let's discuss.

weyrick (changes requested):

agent/rpc_from.go
dscabral: I understand. Where can I get the list of new policies? Every time It receives policies to handle on `handleAgentPolicies`??

agent/rpc_from.go
```diff
@@ -38,8 +28,41 @@ func (a *orbAgent) handleGroupMembership(rpc fleet.GroupMembershipRPCPayload) {
 	}
 }
 
-func (a *orbAgent) handleAgentPolicies(rpc []fleet.AgentPolicyRPCPayload) {
-
+func (a *orbAgent) handleAgentPolicies(rpc []fleet.AgentPolicyRPCPayload, fullList bool) {
+	if fullList {
+		policies, err := a.policyManager.GetRepo().GetAll()
+		if err != nil {
+			a.logger.Error("failed to retrieve policies on handle subscriptions")
```
weyrick, line 35: return after the log i guess, as policies will be nil

agent/rpc_from.go
```diff
@@ -38,8 +28,41 @@ func (a *orbAgent) handleGroupMembership(rpc fleet.GroupMembershipRPCPayload) {
...
+		}
+		// Compare the list of policies the check which one should be removed
+		// After a editing a policy
+		policyRemove := map[string]bool{}
+		for _, p := range policies {
```
weyrick, line 40: this is O(n2) in complexity, not a problem for our small policy counts right now but would be if there were hundreds of policies. it might make sense to make a hash table of rpc.payload.ID => bool before the `range policies` loop and use that instead of the embedded loop

agent/rpc_from.go
```diff
@@ -38,8 +28,41 @@ func (a *orbAgent) handleGroupMembership(rpc fleet.GroupMembershipRPCPayload) {
...
+			}
+		}
+		// Remove only the policy which should be removed
+		for k, v := range policyRemove {
+			policy, err := a.policyManager.GetRepo().Get(k)
+			if err != nil {
+				a.logger.Error("failed to retrieve policy", zap.String("policy_id", k), zap.Error(err))
+			}
+			if v == true {
+				a.policyManager.ManagePolicy(fleet.AgentPolicyRPCPayload{
```
weyrick, line 55: hm I don't like using ManagePolicy here with a mocked AgentPolicyRPCPayload. Let's refactor the logic out of ManagePolicy for "remove" to a common method we can call.

weyrick (changes requested):

agent/policyMgr/manager.go
```diff
@@ -21,6 +21,7 @@ type PolicyManager interface {
 	GetRepo() policies.PolicyRepo
 	ApplyBackendPolicies(be backend.Backend) error
 	RemoveBackendPolicies(be backend.Backend, permanently bool) error
+	RemovePolicy(policyID string, policyName string, beName string)
```
weyrick, line 24: let's have this return `error`

agent/policyMgr/manager.go
```diff
@@ -114,30 +115,34 @@ func (a *policyManager) ManagePolicy(payload fleet.AgentPolicyRPCPayload) {
...
-		err := be.RemovePolicy(pd)
-		if err != nil {
-			a.logger.Warn("policy failed to remove", zap.String("policy_id", payload.ID), zap.String("policy_name", payload.Name), zap.Error(err))
-		}
-		// Remove policy from orb-agent local repo
-		err = a.repo.Remove(pd.ID)
-		if err != nil {
-			a.logger.Warn("policy failed to remove local", zap.String("policy_id", pd.ID), zap.String("policy_name", pd.Name), zap.Error(err))
-		}
+		a.RemovePolicy(payload.ID, payload.Name, payload.Backend)
```
weyrick, line 118: handle err here, log if not nil

agent/policyMgr/manager.go
```diff
@@ -114,30 +115,34 @@ func (a *policyManager) ManagePolicy(payload fleet.AgentPolicyRPCPayload) {
...
 	}
+}
 
+func (a *policyManager) RemovePolicy(policyID string, policyName string, beName string) {
+	var pd = policies.PolicyData{
+		ID:   policyID,
+		Name: policyName,
+	}
+	if !backend.HaveBackend(beName) {
+		a.logger.Warn("policy remove for a backend we do not have, ignoring", zap.String("policy_id", pd.ID), zap.String("policy_name", pd.Name))
```
weyrick, line 131: return new error w/msg instead of log

agent/policyMgr/manager.go
```diff
@@ -114,30 +115,34 @@ func (a *policyManager) ManagePolicy(payload fleet.AgentPolicyRPCPayload) {
...
+		return
+	}
+	be := backend.GetBackend(beName)
+	// Remove policy via http request
+	err := be.RemovePolicy(pd)
+	if err != nil {
+		a.logger.Warn("policy failed to remove", zap.String("policy_id", pd.ID), zap.String("policy_name", pd.Name), zap.Error(err))
```
weyrick, line 138: return err instead of log

agent/policyMgr/manager.go
```diff
@@ -114,30 +115,34 @@ func (a *policyManager) ManagePolicy(payload fleet.AgentPolicyRPCPayload) {
...
+	}
+	// Remove policy from orb-agent local repo
+	err = a.repo.Remove(pd.ID)
+	if err != nil {
+		a.logger.Warn("policy failed to remove local", zap.String("policy_id", pd.ID), zap.String("policy_name", pd.Name), zap.Error(err))
```
weyrick, line 143: return err instead of log

agent/rpc_from.go
```diff
@@ -33,43 +33,36 @@ func (a *orbAgent) handleAgentPolicies(rpc []fleet.AgentPolicyRPCPayload, fullLi
...
-					Name:      policy.Name,
-					Backend:   policy.Backend,
-					Version:   0,
-					Data:      nil,
-				})
+				policy, err := a.policyManager.GetRepo().Get(k)
+				if err != nil {
+					a.logger.Error("failed to retrieve policy", zap.String("policy_id", k), zap.Error(err))
+				}
+				a.policyManager.RemovePolicy(policy.ID, policy.Name, policy.Backend)
```
weyrick, line 55: handle err here

weyrick (changes requested):
just a few more, thanks

weyrick (approved):
nice

PR #873, opened 2022-02-14
test(sinker): add more unit test for sinker parser

mclcavalcante:


dscabral (commented):
The pktvisor output it's only a reference, because we have some extra fields on Sinker. The goal it's to have a test case for each metric.

dscabral (approved):
Looks pretty good. I've not checked if it's covering all cases, but I guess it's covering almost all of them. I've run on my local env and looks good.

weyrick:
don't wait for my review on this please.

PR #884, opened 2022-02-18
Bug/orb agent status

dscabral:
I've created a process to check agents without heartbeats for more than 30 minutes...
I've noticed that when we use docker rm -f, even if we handle the SIGKILL signal on our golang application, the container kills the process before the application handles the signal. I'm researching some ways to work around it. 
```
docker rm -f
The final option for stopping a running container is to use the --force or -f flag in conjunction with the docker rm command. Typically, docker rm is used to remove an already stopped container, but the use of the -f flag will cause it to first issue a SIGKILL.

docker rm ----force foo
If your goal is to erase all traces of a running container, then docker rm -f is the quickest way to achieve that. However, if you want to allow the container to shutdown gracefully you should avoid this option.
```


cmd/agent/main.go
```diff
@@ -81,11 +81,11 @@ func Run(cmd *cobra.Command, args []string) {
 	}
 
 	// handle signals
-	sigs := make(chan os.Signal, 1)
 	done := make(chan bool, 1)
-	signal.Notify(sigs, syscall.SIGINT, syscall.SIGTERM)
 
 	go func() {
+		sigs := make(chan os.Signal, 1)
```
weyrick, line 87: What does moving this inside the go func do for us?

fleet/service.go
```diff
@@ -92,15 +95,19 @@ func (svc fleetService) thing(token, id string, name string, md map[string]inter
 	return thing, nil
 }
 
-func NewFleetService(logger *zap.Logger, auth mainflux.AuthServiceClient, agentRepo AgentRepository, agentGroupRepository AgentGroupRepository, agentComms AgentCommsService, mfsdk mfsdk.SDK) Service {
+func NewFleetService(logger *zap.Logger, auth mainflux.AuthServiceClient, agentRepo AgentRepository, agentGroupRepository AgentGroupRepository, agentComms AgentCommsService, mfsdk mfsdk.SDK, aTicker *time.Ticker, aDone chan bool) Service {
 
-	return &fleetService{
+	service := fleetService{
```
weyrick, line 100: I think we can create aTicket and aDone here, rather than accepting them as arguments

fleet/agent_state_check.go
```diff
@@ -0,0 +1,38 @@
...
+	}
+}
+
+
+func (svc *fleetService) checkAgents() {
+	svc.checkState(time.Now())
+	for {
+		select {
+		case <-svc.aDone:
+			svc.logger.Info("stopping ticker")
```
weyrick, line 32: "stopping stale agent routine"

weyrick (changes requested):

weyrick (approved):

PR #887, opened 2022-02-22
Feat/orb 851 unlink removed sink from dataset and inactivate if in case

mclcavalcante:
- Detaches a sink of all datasets when it is removed.
- If all sinks attached to a dataset are removed, these datasets are invalidated.

### DIFF
--- a/policies/api/http/metrics.go
+++ b/policies/api/http/metrics.go
@@ -376,6 +376,50 @@ func (m metricsMiddleware) ListDatasets(ctx context.Context, token string, pm po
 	return m.svc.ListDatasets(ctx, token, pm)
 }
 
+func (m metricsMiddleware) InactivateDatasetBySinkID(ctx context.Context, sinkID string, token string) error {
+	ownerID, err := m.identify(token)
+	if err != nil {
+		return err
+	}
+
+	defer func(begin time.Time) {
+		labels := []string{
+			"method", "inactivateDatasetByGroupID",
+			"owner_id", ownerID,
+			"policy_id", "",
+			"dataset_id", "",
+		}
+
+		m.counter.With(labels...).Add(1)
+		m.latency.With(labels...).Observe(float64(time.Since(begin).Microseconds()))
+
+	}(time.Now())
+
+	return m.svc.InactivateDatasetBySinkID(ctx, sinkID, token)
+}
+
+func (m metricsMiddleware) DeleteSinkFromDataset(ctx context.Context, sinkID string, token string) error {
+	ownerID, err := m.identify(token)
+	if err != nil {
+		return err
+	}
+
+	defer func(begin time.Time) {
+		labels := []string{
+			"method", "inactivateDatasetByGroupID",
+			"owner_id", ownerID,
+			"policy_id", "",
+			"dataset_id", "",
+		}
+
+		m.counter.With(labels...).Add(1)
+		m.latency.With(labels...).Observe(float64(time.Since(begin).Microseconds()))
+
+	}(time.Now())
+
+	return m.svc.DeleteSinkFromDataset(ctx, sinkID, token)
+}
+
 func (m metricsMiddleware) identify(token string) (string, error) {
 	ctx, cancel := context.WithTimeout(context.Background(), time.Second)
 	defer cancel()
--- a/policies/policies.go
+++ b/policies/policies.go
@@ -102,6 +102,12 @@ type Service interface {
 
 	// ListDatasets retrieve a list of Dataset by owner
 	ListDatasets(ctx context.Context, token string, pm PageMetadata) (PageDataset, error)
+
+	// InactivateDatasetBySinkID inactivate a dataset
+	InactivateDatasetBySinkID(ctx context.Context, sinkID string, token string) error
+
+	// DeleteSinkFromDataset removes a sink from a dataset
+	DeleteSinkFromDataset(ctx context.Context, sinkID string, token string) error
 }
 
 type Repository interface {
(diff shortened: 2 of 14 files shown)


policies/policy_service.go
```diff
@@ -328,3 +328,43 @@ func (s policiesService) ListDatasets(ctx context.Context, token string, pm Page
...
+	ownerID, err := s.identify(token)
+	if err != nil {
+		return err
+	}
+
+	if sinkID == "" {
+		return ErrMalformedEntity
+	}
+
+	datasets, err := s.repo.RetrieveAllDatasetsInternal(ctx, ownerID)
```
dscabral, line 342: You should create a method on the repository to do it based on the provided sink id. I know that you're trying to keep the single responsibility principle ok, but got through each dataset, check if it has only one sink and if the sink it's was removed, looks like a lot of steps to do the same.

policies/redis/consumer/streams.go
```diff
@@ -98,6 +145,20 @@ func (es eventStore) handleAgentGroupRemove(ctx context.Context, groupID string,
 	return nil
 }
 
+func (es eventStore) handleSinkRemove(ctx context.Context, sinkID string, token string) error {
+
```
dscabral, line 149: Maybe first you should delete the sink from the dataset, after that, you should retrieve the dataset and check the size of the sinks it equals zero, and then you inactive the dataset. What do you think?

dscabral (changes requested):

dscabral (approved):

policies/api/http/metrics.go
```diff
@@ -376,6 +398,28 @@ func (m metricsMiddleware) ListDatasets(ctx context.Context, token string, pm po
...
 
+func (m metricsMiddleware) DeleteSinkFromDataset(ctx context.Context, sinkID string, token string) ([]policies.Dataset, error) {
+	ownerID, err := m.identify(token)
+	if err != nil {
+		return []policies.Dataset{}, err
+	}
+
+	defer func(begin time.Time) {
+		labels := []string{
+			"method", "inactivateDatasetByGroupID",
```
weyrick, line 409: method name doesn't match?

policies/policies.go
```diff
@@ -148,4 +154,10 @@ type Repository interface {
 
 	// RetrieveAllDatasetsByOwner retrieves the subset of Datasets owned by the specified user
 	RetrieveAllDatasetsByOwner(ctx context.Context, ownerID string, pm PageMetadata) (PageDataset, error)
+
+	// InactivateDatasetByID inactivate a dataset
+	InactivateDatasetByID(ctx context.Context, sinkID string, ownerID string) error
+
+	// DeleteSinkFromDataset removes a sink from a dataset
+	DeleteSinkFromDataset(ctx context.Context, sinkID string, ownerID string) ([]Dataset, error)
```
weyrick, line 162: This actually removes the sinkID from _all_ Datasets, right? So the function should be called `DeleteSinkFromAllDatasets`

policies/policies.go
```diff
@@ -148,4 +154,10 @@ type Repository interface {
 
 	// RetrieveAllDatasetsByOwner retrieves the subset of Datasets owned by the specified user
 	RetrieveAllDatasetsByOwner(ctx context.Context, ownerID string, pm PageMetadata) (PageDataset, error)
+
+	// InactivateDatasetByID inactivate a dataset
+	InactivateDatasetByID(ctx context.Context, sinkID string, ownerID string) error
```
weyrick, line 159: this should be datasetID, not sinkID, right?

sinks/redis/producer/events.go
```diff
@@ -49,12 +49,14 @@ func (cce createSinkEvent) Encode() map[string]interface{} {
 }
 
 type deleteSinkEvent struct {
-	id string
+	sinkID    string
+	token string
```
weyrick, line 53: we don't want to send tokens through redis events ... we don't need to because the internal events are considered a secure channel. instead we should send whatever information is needed on the other side (ownerID, etc) and call an "Internal" service function which don't take a token (see the existing examples). so we probably want `InactivateDatasetByIDInternal` for example, which will never be called by e.g. the REST API. let me if we need to discuss this to make it clearer.

policies/redis/consumer/events.go
```diff
@@ -15,3 +15,9 @@ type removeAgentGroupEvent struct {
 	token     string
 	timestamp time.Time
 }
+
+type removeSinkEvent struct {
+	sinkID    string
+	token     string
```
weyrick, line 21: see other note on token

weyrick (changes requested):

sinks/redis/producer/events.go
mclcavalcante: Make sense, got it!

policies/redis/producer/streams.go
```diff
@@ -229,17 +229,17 @@ func (e eventStore) ValidatePolicy(ctx context.Context, token string, p policies
...
-func (e eventStore) InactivateDatasetByID(ctx context.Context, datasetID string, token string) error {
-	ds, err := e.svc.ViewDatasetByID(ctx, token, datasetID)
+func (e eventStore) InactivateDatasetByIDInternal(ctx context.Context, datasetID string, ownerID string) error {
+	ds, err := e.svc.ViewDatasetByIDInternal(ctx, ownerID, datasetID)
 	if err != nil {
 		return err
 	}
 
-	if err := e.svc.InactivateDatasetByID(ctx, datasetID, token); err != nil {
+	if err := e.svc.InactivateDatasetByIDInternal(ctx, datasetID, ownerID); err != nil {
```
weyrick, line 242: we see now that ViewDatasetByIDInternal takes (ctx, ownerID, datasetID) but InactivateDatasetByIDInternal takes (ctx, datasetID, ownerID)... we should instead keep the order of the arguments consistent, which makes for a nicer API to develop with helps avoid programmer error (because they are both strings, they are easily swapped). since ViewDatasetByIDInternal already exists, let's change InactivateDatasetByIDInternal to match

weyrick (changes requested):

policies/redis/producer/streams.go
mclcavalcante: oh I see, ok, changing it

weyrick (approved):

PR #888, opened 2022-02-22
fix(orb-ui): ironclad email issues

gpazuch:


ui/angular.json
```diff
@@ -104,7 +104,11 @@
         "serve": {
           "builder": "@angular-devkit/build-angular:dev-server",
           "options": {
-            "browserTarget": "orb-ui:build"
+            "browserTarget": "orb-ui:build",
+            "ssl": true,
+            "sslKey": "scripts/key.pem",
+            "sslCert": "scripts/cert.pem",
+            "disableHostCheck": true
```
gpazuch, line 111: @iatdaitan I think these changes shouldn't be available, and I have created a separate gist for it [here](https://gist.github.com/gpazuch/dd853646b4554fb2d8c9a6fa30ce4479).

In order to use it in local environment for development/testing purposes, one must create a certificate - in this case, placed inside `scripts` folder

```openssl req -newkey rsa:2048 -new -nodes -x509 -days 3650 -keyout key.pem -out cert.pem```

as well as have the IC keys available in the host env.

```
export PS_GROUP_KEY=my-group-key
export PS_SID=my-site-id
```

before running `yarn start` to serve the application.

@manrodrigues with this you should be able to test ironclad locally.

File: ui/src/app/pages/register/register.component.html [hunk 2]
```diff
@@ -156,72 +156,74 @@ <h1 id="title" class="title">Create an account</h1>
...
           </ng-container>
         </div>
 
         <ps-click-wrap
-          *ngIf="_isProduction && _psEnabled"
-          accessId="{{_sid}}"
-          groupKey="{{_groupKey}}"
-          signerIdSelector="input-name"
-          confirmationEmail="1">
+          *ngIf="_psEnabled"
```
gpazuch, line 191: @iatdaitan I needed to change the signer ID to email, in order to be able to send the email information, since `custom_data` param. only holds first_name, last_name and company_name; see [this page for ref](https://clickwrap-developer.ironcladapp.com/reference/url-parameters-for-send)

ui/src/app/pages/register/register.component.ts
```diff
@@ -88,12 +89,17 @@ export class RegisterComponent extends NbRegisterComponent implements OnInit {
         company,
       }).subscribe(
         respReg => {
+          const first_name = this.user.fullname.split(' ')[0];
+          const last_name = this.user.fullName.replace(`${first_name} `, '');
+          _ps.getByKey(this._groupKey).send('updated', { custom_data: { first_name, last_name } });
```
gpazuch, line 94: @iatdaitan Instead of disabling auto sending contract acceptance, I update user info upon register.

Optionally, I can disable all events and manually call the endpoint for acceptance at this point.

ui/src/app/pages/register/register.component.ts
```diff
@@ -88,12 +89,17 @@ export class RegisterComponent extends NbRegisterComponent implements OnInit {
...
+
           this.submitted = false;
 
           if (respReg.isSuccess()) {
```
gpazuch, line 98: @iatdaitan outside of this issue's scope, register errors such as name or email already in use are not being correctly treated in this page, leading to a locked state where <Register> button no longer enables it self, even after changing inputs.

ui/angular.json
iatdaitan: Thank you.

ui/src/app/pages/register/register.component.html
iatdaitan: Got it, thank you.

ui/src/app/pages/register/register.component.ts
iatdaitan: Nice change.

ui/src/app/pages/register/register.component.ts
iatdaitan: @manrodrigues please, consider it during your test. I am going to open an issue to deal with those error enhancements.

iatdaitan (approved):

PR #893, opened 2022-02-25
fix(orb-ui): agent group matching agents list issues

gpazuch:
image
image

1920x1080 & 1366x768

*fix issue update matching tags breaks table column configuration

*fix issue update matching tags breaks rows - change detection not enabled.


gpazuch:
@iatdaitan Can we have a lower priority task for fluid column widths?
I focused on the update matches issue, and just a small touch on the columns widths

iatdaitan (approved):

PR #902, opened 2022-02-25
Bug/orb fleet fail subscribe only one group

dscabral:
Fix subscription problem when the agent has no match groups

### DIFF
--- a/agent/rpc_from.go
+++ b/agent/rpc_from.go
@@ -63,7 +63,7 @@ func (a *orbAgent) handleAgentPolicies(rpc []fleet.AgentPolicyRPCPayload, fullLi
 	}
 
 	for _, payload := range rpc {
-		if payload.Action != "" {
+		if payload.Action != "sanitize" {
 			a.policyManager.ManagePolicy(payload)
 		}
 	}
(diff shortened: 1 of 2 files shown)


fleet/comms.go
```diff
@@ -174,29 +169,35 @@ func (svc fleetCommsService) NotifyAgentAllDatasets(a Agent) error {
...
-			Backend:   policy.Backend,
-			Version:   policy.Version,
-			Data:      pdata,
-			DatasetID: policy.DatasetId,
 		}
-
+	} else {
+		// Even with no policies, we should send the signal to agent for policy sanitization
+		payload = make([]AgentPolicyRPCPayload, 1)
+		payload[0] = AgentPolicyRPCPayload{}
```
weyrick, line 200: let's fill out `Action` here to be `sanitize` or `cleanup` or something

agent/rpc_from.go
```diff
@@ -18,14 +18,14 @@ func (a *orbAgent) handleGroupMembership(rpc fleet.GroupMembershipRPCPayload) {
...
 		a.subscribeGroupChannels(rpc.Groups)
+		err := a.sendAgentPoliciesReq()
+		if err != nil {
+			a.logger.Error("failed to send agent policies request", zap.Error(err))
+		}
 	} else {
 		// otherwise, just add these subscriptions to the existing list
 		a.subscribeGroupChannels(rpc.Groups)
 	}
-	err := a.sendAgentPoliciesReq()
```
weyrick, line 25: Doesn't moving this mean we will not ask for policies unless we get the FullList of groups, and if so is that right?

weyrick (changes requested):

weyrick (approved):

PR #913, opened 2022-03-03
fix(orb-ui): Datatable pagination and sorting issues

gpazuch:
* filter on client side

* sort client side

* tag filtering by key and value - try multiple keywords

* dynamic pagination and virtual scroll enabled

* table height and width with better fluidness

* some columns now have tooltips

* all services now offer convenience method to query and cache all pages of data
** agents, agent groups, sinks, datasets and policies.

* css improvements on table elements and table pages

* code style fixes


gpazuch:
@iatdaitan 
it remains to be improved/done

* fluid column witdh
* better solution for tags display => maybe css animation on hover to scroll tags list in place
* cache and refresh strategy

iatdaitan (approved):

PR #915, opened 2022-03-03
change view to get mix of agents

dscabral:
Change query on view creation (agent_group_membership) to create a row for agents which has a mix of orb_tags and agent_tags

### DIFF
--- a/fleet/postgres/agents.go
+++ b/fleet/postgres/agents.go
@@ -315,7 +315,10 @@ func (r agentRepository) RetrieveByIDWithChannel(ctx context.Context, thingID st
 }
 
 func (r agentRepository) Save(ctx context.Context, agent fleet.Agent) error {
-
+	tx, err := r.db.BeginTxx(ctx, nil)
+	if err != nil {
+		return err
+	}
 	q := `INSERT INTO agents (name, mf_thing_id, mf_owner_id, mf_channel_id, orb_tags, agent_tags, agent_metadata, state)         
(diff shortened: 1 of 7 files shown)


fleet/postgres/init.go
```diff
@@ -82,7 +82,7 @@ func migrateDB(db *sqlx.DB) error {
 					)`,
 					`CREATE INDEX ON agent_groups (mf_owner_id)`,
 					`CREATE INDEX ON agent_groups USING gin (tags)`,
-					`CREATE VIEW agent_group_membership(agent_groups_id, agent_groups_name, agent_mf_thing_id, agent_mf_channel_id, group_mf_channel_id, mf_owner_id, agent_state) as
+					`CREATE or REPLACE VIEW agent_group_membership(agent_groups_id, agent_groups_name, agent_mf_thing_id, agent_mf_channel_id, group_mf_channel_id, mf_owner_id, agent_state) as
```
weyrick, line 85: No, as I mentioned in slack, any [DDL](https://www.postgresql.org/docs/14/ddl.html) requires a migration

fleet/postgres/agents.go
```diff
@@ -315,7 +315,10 @@ func (r agentRepository) RetrieveByIDWithChannel(ctx context.Context, thingID st
 }
 
 func (r agentRepository) Save(ctx context.Context, agent fleet.Agent) error {
-
+	tx, err := r.db.BeginTxx(ctx, nil)
```
weyrick, line 318: Since there is only one statement here, what does the transaction do for us?

weyrick (changes requested):

weyrick (approved):

PR #942, opened 2022-03-10
feature(policy): add support for yaml formatted policy

mclcavalcante:


dscabral (changes requested):
You should update the open API spec as well.
Despite it, it's pretty good.

dscabral:
Thinking better, we still have some work to do.
Since we're creating a new database field to receive the policy YAML, we should send the value through the application to the agent. The Dataset creation and the Editing Policy should notify the Fleet to retrieve the YAML and send it in an RPC call to the agent. Probably you have to change the `handleDatasetCreate` method on Fleet service and the `NotifyGroupNewDataset` needs some change on `proto definition`, creating a new field to the policy data.
`p, err := svc.policyClient.RetrievePolicy(ctx, &pb.PolicyByIDReq{PolicyID: policyID, OwnerID: ownerID})` the return of this gRPC call needs to return a new field.

dscabral (changes requested):
Thinking better, we still have some work to do.
Since we're creating a new database field to receive the policy YAML, we should send the value through the application to the agent. The Dataset creation and the Editing Policy should notify the Fleet to retrieve the YAML and send it in an RPC call to the agent. Probably you have to change the handleDatasetCreate method on Fleet service and the NotifyGroupNewDataset needs some change on proto definition, creating a new field to the policy data.
p, err := svc.policyClient.RetrievePolicy(ctx, &pb.PolicyByIDReq{PolicyID: policyID, OwnerID: ownerID}) the return of this gRPC call needs to return a new field.

weyrick (changes requested):
this looks good but i don't see a change to return the new PolicyData and Format fields in the REST API responses, so it's not accessible, is it?

mclcavalcante:
> this looks good but i don't see a change to return the new PolicyData and Format fields in the REST API responses, so it's not accessible, is it?

@weyrick It is only returned by policy view endpoint, do you think it is important to have it accessible by other responses? @gpazuch said he needs to be able to get it by calling policy view, but I could definitely add to other API responses too

gpazuch:
@mclcavalcante I didn't get the context of the mention above.

If possible, police_data and format should be returning with all requests, although orb-ui is only using this information in the Policy Create/Policy Edit pages.

I understand orb-ui will need to update some views to display the yaml information when relevant, for the policy, since the backend will no longer convert the yaml to json - I imagine this issue being relevant on agent-view or agent-policy-views maybe.

I'd also like to point out that the agent-view issue of this sprint [check PR here](https://github.com/ns1labs/orb/pull/968) - although bad design from the front end side, was triggered because the policy information listed in the list-policies end point differs from the full policy that can be retrieved from get-policy-by-id endpoint, which contains the agent_metadata and last_hb_data fields.

weyrick (approved):

dscabral (approved):

PR #963, opened 2022-03-16
automated tests for sink removal

manrodrigues:


python-test/features/integration.feature
```diff
@@ -147,7 +147,7 @@ Scenario: Sink with invalid endpoint
         And referred agent is subscribed to a group
         And that a sink with invalid endpoint already exists
         And that a policy using: handler=dns, description='policy_dns', host_specification=10.0.1.0/24,10.0.2.1/32,2001:db8::/64, bpf_filter_expression=udp port 53, pcap_source=libpcap, only_qname_suffix=[.foo.com/ .example.com], only_rcode=2 already exists
-    When a new dataset is created using referred group, sink and policy ID
+    When a new dataset is created using referred group, policy and 1 sink ID
```
leoparente, line 150: should not be ".. group, policy ID and 1 sink "?

python-test/features/steps/control_plane_policies.py
```diff
@@ -281,7 +281,7 @@ def apply_n_policies_x_times(context, amount_of_policies, type_of_policies, amou
         create_new_policy(context, args_for_policies[n][1])
         check_policies(context)
         for x in range(int(amount_of_datasets)):
-            create_new_dataset(context)
+            create_new_dataset(context, 1, 'sink')
```
leoparente, line 284: this '1' value should not be inserted by the @step?

python-test/features/integration.feature
leoparente, line 150: It should not be "... group, policy ID and 1 sink"? 
or the way it was" ... group, 1 sink and policy ID"

manrodrigues (commented):
I have removed ID, because I agree it was confuse and there is no necessity to specify it (all elements are linked through ID)

Sorry, @leoparente , I closed your comments by mistake.

About the hardcoded part, on those scenarios, is always 1 sink... and we don't have options on step description

leoparente (approved):

PR #965, opened 2022-03-16
Feat/orb sinker redis

dscabral:


sinker/heartbeats.go
```diff
@@ -22,7 +22,7 @@ const (
 
 func (svc *sinkerService) sendSingleHeartbeat(t time.Time) {
 	svc.logger.Info("heartbeat")
```
weyrick, line 24: this is very noisy in our logs and i don't think provides any value - let's remove while we're working on this

sinker/heartbeats.go
```diff
@@ -22,7 +22,7 @@ const (
 
 func (svc *sinkerService) sendSingleHeartbeat(t time.Time) {
```
weyrick, line 23: I don't remember/understand the heartbeating here. I understand it's a way to try to track idle Sinks... but why is it sending redis events? And does it actually work, when the sinker service scales horizontally to multiple services?

Should we take the time now to change this to the way Agents work - a database field tracking last write, and a task which updates to idle based on that?

sinker/redis/sinker.go
```diff
@@ -0,0 +1,99 @@
...
+	for iter.Next(context.Background()) {
+		cfg, err := s.Get(strings.TrimPrefix(iter.Val(), fmt.Sprintf("%s-", keyPrefix)))
+		if err != nil {
+			s.logger.Error("failed to retrieve config", zap.Error(err))
+			continue
+		}
+		configs = append(configs, cfg)
+	}
+	if err := iter.Err(); err != nil {
+		panic(err)
```
weyrick, line 95: we should never panic like this unless we can not recover

sinker/redis/sinker.go
```diff
@@ -0,0 +1,99 @@
...
+		return err
+	}
+	if err := s.Add(config); err != nil {
+		return err
+	}
+	return nil
+}
+
+func (s *sinkerCache) GetAll() ([]config.SinkConfig, error) {
+	iter := s.client.Scan(context.Background(), 0, fmt.Sprintf("%s-*",keyPrefix), 0).Iterator()
```
weyrick, line 84: this doesn't support multi-tenancy though, does it? we would need the prefix to include ownerID

weyrick (changes requested):

sinker/config/repo.go
```diff
@@ -5,10 +5,10 @@
 package config
 
 type ConfigRepo interface {
-	Exists(sinkID string) bool
+	Exists(ownerID string, sinkID string) bool
 	Add(config SinkConfig) error
-	Remove(sinkID string) error
-	Get(sinkID string) (SinkConfig, error)
+	Remove(ownerID, sinkID string) error
```
weyrick, line 10: `ownerID string`  ... (does it assume string if missing?)

weyrick (changes requested):

sinker/config_state_check.go
```diff
@@ -0,0 +1,56 @@
...
+
+	for _, ownerID := range owners {
+		configs, err := svc.sinkerCache.GetAll(ownerID)
+		if err != nil {
+			svc.logger.Error("unable to retrieve policy state", zap.Error(err))
+			return
+		}
+		for _, cfg := range configs {
+			// Set idle if the sinker is more then 30 minutes not been sending metrics
+			if cfg.LastRemoteWrite.Add(DefaultTimeout).Before(time.Now()) {
```
weyrick, line 35: So this doesn't work (well) because every instance of sinker in the cluster will be duplicating the timeout calculations and updates. The reason it works better for Agents is because the "last timestamp" state and the action for setting "stale" status is done in postgres - all instances in the cluster will just call the function to update the status in the database, and only the first one will actually do anything. In this code, all instances will all write to redis.  

I think the better way to do this is to do what we do with Agents, and add a new 'ts_last_sink' field to postgres which is updated upon successful sink. Then this function can do what agent does, and call the repo function to update all 'active' sinks with a 'ts_last_sink' older than DefaultTimeout to 'idle'

sinker/config_state_check.go
```diff
@@ -0,0 +1,56 @@
...
+		return
+	}
+
+	for _, ownerID := range owners {
+		configs, err := svc.sinkerCache.GetAll(ownerID)
+		if err != nil {
+			svc.logger.Error("unable to retrieve policy state", zap.Error(err))
+			return
+		}
+		for _, cfg := range configs {
```
weyrick, line 33: it should only be setting timers for cfg.State which are "active"

weyrick (changes requested):

sinker/config_state_check.go
dscabral: ts_last_sink should be a field from what database? I'm using lastRemoteWrite, which is a timestamp field, stored on Redis to check the last sink time. I guess it's working the same way the ts_last_sink would work on Postgres. I've changed the function to only set idle for active configurations, so the first one is going the change the value to idle if we have more than one instance. I'm asking all this because we don't have a Postgres instance for sinker service and I'm using Redis to store the configs...

weyrick: Yeah it would be the sinks database/table so this "state check" functionality would have to move there instead of sinker. I do see there would be some complication in keeping that up to date when sinker writes metrics - it would have to go through the event stream.

Ok if we leave this code in place - I'm still concerned that all instances will all iterate through all sinks in redis every couple of minutes. How about if instead of changing the state to 'idle' in redis, we removed it, so that we _only_ store "active" sinks in redis? Presumably when that sink starts receiving metrics again, it updates itself to "active" and recaches?

sinker/config_state_check.go
```diff
@@ -0,0 +1,57 @@
...
+			// Set idle if the sinker is more then 30 minutes not sending metrics (Remove from Redis)
+			if cfg.LastRemoteWrite.Add(DefaultTimeout).Before(time.Now()) {
+				if cfg.State == config.Active {
```
weyrick, line 36: i think this will always be true right?

sinker/config_state_check.go
```diff
@@ -0,0 +1,57 @@
...
+					if err := svc.sinkerCache.Remove(cfg.OwnerID, cfg.SinkID); err != nil {
```
weyrick, line 37: did you confirm this won't fail if it's already been removed? we know that the different sinker nodes will race to do this, and only one will succeed

dscabral: If Redis try to remove a non-existing key, it's ignored.

weyrick (approved):

PR #968, opened 2022-03-22
fix(orb-ui): agent view issues

gpazuch:
Refactor agent view page into smaller components for each card/information section.
This should improve maintenance and readability.
It also makes easier to split the chain of observables and manipulate the mappings of the agent policy state to inner units.


gpazuch:
@iatdaitan can you create new sprint and release for this sprint?

Also, would you please review this refactor? I've started to refactor this page into what it should, I believe, since it was buggy and very messy to deal with the chained observables I needed to retrieve.

iatdaitan (approved):
Nive work @gpazuch .

PR #970, opened 2022-03-22
fix(orb-ui): fix wrong text on agent create provisioning command view.

gpazuch:


ui/src/app/pages/fleet/agents/key/agent.key.component.html
```diff
@@ -28,10 +28,6 @@
       </a>
       for more advanced options.
     </p>
-    <p>
-      To use the provisioning command, you must replace AGENT_KEY with the Agent Key given to you during agent
-      creation. If you have lost the Agent Key, you must remove and create a new agent.
-    </p>
     <p>
       If you are looking for scrapping real data from your Agent, please consider changing the mock with your Agent default iface.
     </p>
```
manrodrigues, line 33: `If you are looking for scrapping real data from your Agent, please consider changing the mock with your Agent default iface.` this text is not necessary once in this modal we already have this ```     <p>
      Replace <b style="color:#df316f">mock</b>
      interface with host
      network interface. <u nbTooltip="e.g. eth0, wlan0" style="color: #969fb9">?</u>
      <br>
      <a href="https://getorb.io/docs/#running-orb-agent"
         target="_blank">See here ```.

I guess the best option is to use a mixed of them:

I mean:

Replace mock .... if you are looking for scrapping real data..

What do you think?

gpazuch: @manrodrigues I went right past that. ty :+1:

manrodrigues (approved):

PR #974, opened 2022-03-22
Feat/orb sinks validate prom config

dscabral:
Create Prometheus client
Validate Prometheus configuration on creating and editing process
Handle sinks states (sink and sinker)


weyrick (changes requested):
comments in slack

dscabral:
After a review, this approach it's not needed anymore.
It's breaking some architectural concepts and doesn't make sense to keep it.
It's important to keep Sink agnostic, regarding the technology used to store metrics and keep the concrete implementations on Sinker.

PR #978, opened 2022-03-24
Feat/orb fleet matching groups

dscabral:


fleet/api/http/openapi.yaml
```diff
@@ -742,6 +742,28 @@ components:
           type: string
           format: uuid
           description: Communication channel ID (UUIDv4), unique to this agent and created at agent creation
+        matching_groups:
+          type: object
+          description: Groups currently matching this agent
```
weyrick, line 747: Groups this Agent currently belongs to, according to matching agent and group tags

fleet/api/http/openapi.yaml
```diff
@@ -742,6 +742,28 @@ components:
...
+                properties:
+                  group_id:
+                    type: string
+                    format: uuid
+                    description: group id
+                  name:
+                    type: string
+                    description: group name
+                    example: 'group-1'
+                  channel:
```
weyrick, line 763: Let's remove channel: I'd like to keep this out of the public interface, we don't return it in the group endpoints

fleet/api/http/openapi.yaml
```diff
@@ -742,6 +742,28 @@ components:
...
           description: Communication channel ID (UUIDv4), unique to this agent and created at agent creation
+        matching_groups:
+          type: object
+          description: Groups currently matching this agent
+          properties:
+            groups:
+              type: array
+              items:
+                type: object
+                description: total agents matching
```
weyrick, line 753: from copy/paste?

fleet/postgres/agents.go
```diff
@@ -391,19 +391,22 @@ func (r agentRepository) RetrieveByID(ctx context.Context, ownerID string, thing
...
+			ts_last_hb,
+			json_build_object('groups', json_strip_nulls(json_agg(json_build_object('group_id', agm.agent_groups_id, 'name', agm.agent_groups_name, 'channel', agm.group_mf_channel_id)))) matching_groups
 		FROM agents 
+			LEFT JOIN agent_group_membership agm on agents.mf_owner_id = agm.mf_owner_id and agents.mf_thing_id = agm.agent_mf_thing_id
 		WHERE 
-			mf_thing_id = $1 
-			AND mf_owner_id = $2;`
+		mf_thing_id = $1 
+		AND agents.mf_owner_id = $2
+		group by mf_thing_id, name, agents.mf_owner_id, mf_channel_id, ts_created, orb_tags, agent_tags, agent_metadata, state, last_hb_data, ts_last_hb;`
```
weyrick, line 409: The need for this "group by" seems really weird to me - we will have to list all the columns? When a new column is added, we will have to remember to add it here to?

Maybe it makes sense to make this a separate query, rather than a join. In fact, if we go further and make it an entirely separate (new) endpoint, then the currently GET agent is unaffected, and the UI can make a concurrent request, possibly improving efficiency.

weyrick (changes requested):

weyrick (approved):

PR #1002, opened 2022-03-28
orb(ui): fix dataset details modal failed state

gpazuch:
* add message for non-retrievable entities - agent group, agent policy, sink;
* fix requests failing on error since forkjoin quits on error - adjust services accordingly


gpazuch:
image

gpazuch:
image

PR #1008, opened 2022-03-28
Feature/885 ability duplicate existing policy

mclcavalcante:
Add endpoint /policies/agent/{id}/duplicate which sends in the request the policy ID that is wanted to duplicate and return a new identical policy, differentiating by name and ID


policies/policy_service.go
```diff
@@ -357,4 +358,33 @@ func (s policiesService) InactivateDatasetByIDInternal(ctx context.Context, owne
 	}
 
 	return nil
+}
+
+func (s policiesService) DuplicatePolicy(ctx context.Context, token string, policyID string) (Policy, error) {
+
```
dscabral, line 364: I guees it should save duplicated policy on database. Probably receive at least the new name from the client or create a logic to generate an automatic name.

dscabral (changes requested):
We still have some work to do. =)

policies/mocks/policies.go
```diff
@@ -32,6 +32,19 @@ func (m *mockPoliciesRepository) ActivateDatasetByID(ctx context.Context, datase
...
 
+func (m *mockPoliciesRepository) RetrieveAllPoliciesInternal(ctx context.Context, ownerID string) ([]policies.Policy, error) {
+	var policyList []policies.Policy
+	id := uint64(0)
+	for _, p := range m.pdb {
+		if p.MFOwnerID == ownerID {
+			policyList = append(policyList, p)
+		}
+		id++
+	}
```
lpegoraro, line 43: here I don't understand why increment the `var id` if not append `policyList`, also the `var id` is not used for reading anymore, I think we can remove it.

policies/policy_service_test.go
```diff
@@ -1297,6 +1297,69 @@ func testSortPolicies(t *testing.T, pm policies.PageMetadata, ags []policies.Pol
...
+	for desc, tc := range cases {
+		t.Run(desc, func(t *testing.T) {
+			policyDuplicated, err := svc.DuplicatePolicy(context.Background(), tc.token, tc.policyID)
+			assert.True(t, errors.Contains(err, tc.err), fmt.Sprintf("%s: expected %s got %s", desc, err, tc.err))
+
+			originalPolicy, _ := svc.ViewPolicyByID(context.Background(), token, tc.policyID)
+			if err == nil {
+				assert.Equal(t, originalPolicy.Policy, policyDuplicated.Policy, fmt.Sprintf("%s: expected %v got %v", desc, originalPolicy, policyDuplicated))
+				assert.Equal(t, tc.expectedName, policyDuplicated.Name.String(),  fmt.Sprintf("%s: expected %v got %v", desc, tc.expectedName, policyDuplicated.Name))
+			}
```
lpegoraro, line 1358: we only validate name here, we could add the most important attrs here to validate the duplicity

lpegoraro (changes requested):
Just a few comments, good job.

lpegoraro (approved):
LGTM

policies/api/http/endpoint.go
```diff
@@ -415,3 +415,31 @@ func listDatasetEndpoint(svc policies.Service) endpoint.Endpoint {
 		return res, nil
 	}
 }
+
+func duplicatePolicyEndpoint(svc policies.Service) endpoint.Endpoint {
+	return func(ctx context.Context, request interface{}) (interface{}, error) {
+		req := request.(viewResourceReq)
```
weyrick, line 421: This should optionally take the new policy name, which if it's passed should be used instead of generating a new name based on the old one

policies/policy_service.go
```diff
@@ -420,4 +422,59 @@ func (s policiesService) DeleteAgentGroupFromAllDatasets(ctx context.Context, gr
...
+		nameSuffix = fmt.Sprintf("_copy")
+	}
+
+	name, err := types.NewIdentifier(existingPolicy.Name.String() + nameSuffix)
+	if err != nil {
+		return Policy{}, err
+	}
+
+	policy := existingPolicy
+	policy.Name = name
```
weyrick, line 471: we have to make sure Version field goes back to 0

policies/policies.go
```diff
@@ -171,4 +174,7 @@ type Repository interface {
 
 	// DeleteAgentGroupFromAllDatasets removes agent group from a dataset
 	DeleteAgentGroupFromAllDatasets(ctx context.Context, groupID string, ownerID string) error
+
+	// RetrieveAllPoliciesInternal retrieves the subset of Policies owned by the specified user
+	RetrieveAllPoliciesInternal(ctx context.Context, ownerID string) ([]Policy, error)
```
weyrick, line 179: I don't think we need this, see below

policies/policy_service.go
```diff
@@ -420,4 +422,59 @@ func (s policiesService) DeleteAgentGroupFromAllDatasets(ctx context.Context, gr
...
+	if err != nil {
+		return Policy{}, err
+	}
+
+	policyList, err := s.repo.RetrieveAllPoliciesInternal(ctx, mfOwnerID)
+	if err != nil {
+		return Policy{}, err
+	}
+
+	nameRegex := fmt.Sprintf("%s_copy", existingPolicy.Name.String())
```
weyrick, line 444: As mentioned, this should only happen if a new name was not passed in

policies/policy_service.go
```diff
@@ -420,4 +422,59 @@ func (s policiesService) DeleteAgentGroupFromAllDatasets(ctx context.Context, gr
...
+	if err != nil {
+		return Policy{}, err
+	}
+
+	existingPolicy, err := s.repo.RetrievePolicyByID(ctx, policyID, mfOwnerID)
+	if err != nil {
+		return Policy{}, err
+	}
+
+	policyList, err := s.repo.RetrieveAllPoliciesInternal(ctx, mfOwnerID)
```
weyrick, line 439: Two issues: one, I'm not sure it's the best strategy to get all policies and try to figure out if the name exists or not. First, seems like a lot of work (what if there are thousand policies?) and second, this code can never be sure there isn't a race and a policy with the same name isn't inserted between the time RetrieveAllPoliciesInternal and SavePolicy is called, so it could still fail.

Even if we use this strategy, don't see why we need RetrieveAllPoliciesInternal since RetrievalAll seems to do what we want already.

I think an alternative strategy would be to not check all policies, and just make three attempts to generate a name: "OLDNAME-Copy", "OLDNAME-Copy2" and "OLDNAME-Copy3". It can try to SavePolicy the first, if it fails try the second, if it fails try the third, if it fails then fail the call.

weyrick (changes requested):

policies/policy_service.go
```diff
@@ -422,3 +423,69 @@ func (s policiesService) DeleteAgentGroupFromAllDatasets(ctx context.Context, gr
...
+		for {
+			policyName, err := types.NewIdentifier(existingPolicy.Name.String() + nameSuffix)
+			if err != nil {
+				return Policy{}, err
+			}
+
+			policy.Name = policyName
+			id, errCreate = s.repo.SavePolicy(ctx, policy)
+			if errCreate != nil {
+				if i < 3{
```
lpegoraro, line 473:
```suggestion
			if errCreate != nil && errCreate == errors.ErrConflict {
				if i < 3 {
```

policies/policy_service.go
```diff
@@ -422,3 +423,69 @@ func (s policiesService) DeleteAgentGroupFromAllDatasets(ctx context.Context, gr
...
+					nameSuffix = fmt.Sprintf("_copy%d", i)
+					continue
+				} else {
+					//if already exists policies named with those three options then call fails
+					return Policy{}, errors.Wrap(errors.New("limit of copies of a single policy exceeded"), errors.ErrConflict)
+				}
+			}
+			break
+		}
+		if errCreate != nil{
```
lpegoraro, line 484:
```suggestion
		if errCreate != nil {
```

policies/policy_service.go
```diff
@@ -422,3 +423,69 @@ func (s policiesService) DeleteAgentGroupFromAllDatasets(ctx context.Context, gr
...
+			return Policy{}, err
+		}
+	}
+	policy.ID = id
+
+	return policy, nil
+}
```
lpegoraro, line 491: Line at end of file

lpegoraro (changes requested):
just a few comments

lpegoraro (approved):
Very well!

policies/policy_service.go
lpegoraro: This was solved in the latter commit.

lpegoraro:
@weyrick I did the review with @mclcavalcante, everything seems ok with the last comment you asked for changes, what do you think?

weyrick (approved):
👍

Issue #957, opened 2022-03-14
All pages: limit 100 propagated by UI


manrodrigues:
Although we have a limit=100 on API response, UI should be able to manage it and display all the existent elements.

check that total policies = 116, but only 100 are being displayed on orb ui:

![image.png](https://images.zenhubusercontent.com/61aa14a937730b887dff82b3/e8e6e2bf-d784-4bf5-9112-0ff29782e659)

![image.png](https://images.zenhubusercontent.com/61aa14a937730b887dff82b3/fb1f7278-3d17-4d13-8837-48762856fb63)

Datasets:

![image.png](https://images.zenhubusercontent.com/61aa14a937730b887dff82b3/d349151a-dfb0-462e-ac85-9d10e6d8051a)

It happens for all orb pages (agents, groups, sink, policies, datasets)

manrodrigues:
ok.. last commit - 7458be83a7f418198931ebc737384e0b60a63ccc

![image](https://user-images.githubusercontent.com/78241475/161600770-8d0a0829-5afc-4c36-ab8b-20d2497f280f.png)

![image](https://user-images.githubusercontent.com/78241475/161600805-151c91bd-8db9-412d-9105-186717edca42.png)

PR #1013, opened 2022-03-30
fix(orb-ui): #957 orb pagination

gpazuch:
* Fix faulty pagination logic, leading to 'extra' items in the list
  * agents service
  * agent groups service
  * agent policies service
  * datasets service
  * sinks service


gpazuch:
@iatdaitan currently, the getAll<Entity> method is implemented with the use of the `expand` rxjs operator.

This means the user is forced to wait for all queries(page queries, since these services are limited to a 100 items per request, thus paginated) to complete in the advent of multiple pages - i.e. 1000 items / 100 items per page => 10 requests.

Considering the service should enforce a debounce time of >250ms per request, it would sum up to 2.5s 'load' time.

I think this could be improved with the use of mergeScan and some arrow functions. 

Perhaps I could have a go on this issue for the next sprint.

manrodrigues (approved):

Issue #746, opened 2022-01-11
Datasets: dataset details fails when a policy, group or sink linked is removed


manrodrigues:
We have a problem when a policy, group or sink linked to dataset is removed:

- span error "failed to fetch"
- All elements after is not displayed (for example, if the group is removed, group, policy and sink is not displayed anymore. If the policy is removed, policy and sink is not displayed anymore).
<br>

Without policy:
<br>

![dataset-policy-deleted](https://user-images.githubusercontent.com/78241475/148988678-9465d85b-778e-422e-978b-51444d5e60ee.gif)

Without group:
<br>
![dataset-group-deleted](https://user-images.githubusercontent.com/78241475/148988736-1e452500-d590-4ea4-8f11-7669f0ef8b0b.gif)

manrodrigues:
Related with #692 and #729

manrodrigues:
Still failing when policy or group is removed:

groups:

![dataset_failed_on_group](https://user-images.githubusercontent.com/78241475/160933786-55a252b7-7d04-48bc-8777-08614fc9876f.gif)

policies:

![dataset_failed_on_policy](https://user-images.githubusercontent.com/78241475/160933816-f167415d-34c2-45ef-8960-25054fda0587.gif)

for sinks removal is ok:

![dataset_ok_without_sink](https://user-images.githubusercontent.com/78241475/160934081-bbf11394-669b-4d39-a4ab-a432c1be202f.gif)

manrodrigues:
ok 


![image](https://user-images.githubusercontent.com/78241475/162480391-b88e9ea5-95fa-49cc-86d7-76c0a3d6ed1a.png)

manrodrigues:
ok 

![image](https://user-images.githubusercontent.com/78241475/162480456-fc6b5e5f-0df3-4abc-acac-e3d2aaca013c.png)

PR #1023, opened 2022-04-01
fix(orb-ui): improve failed to fetch message

gpazuch:
image
#746

### DIFF
--- a/ui/src/app/pages/datasets/details/dataset.details.component.html
+++ b/ui/src/app/pages/datasets/details/dataset.details.component.html
@@ -25,14 +25,19 @@
       <div class="col-6 d-flex flex-column align-items-start">
         <p class="detail-title">Agent Group Name<span class="ns1red">*</span></p>
         <p>{{ agentGroup?.name }}</p>
-        <p *ngIf="errors?.agentGroup" class="detail-title">{{ errors.agentGroup }}<span class="ns1red">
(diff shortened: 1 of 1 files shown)


ui/src/app/pages/datasets/details/dataset.details.component.html
```diff
@@ -25,14 +25,19 @@
       <div class="col-6 d-flex flex-column align-items-start">
         <p class="detail-title">Agent Group Name<span class="ns1red">*</span></p>
         <p>{{ agentGroup?.name }}</p>
-        <p *ngIf="errors?.agentGroup" class="detail-title">{{ errors.agentGroup }}<span class="ns1red">
-          {{ this.dataset.agent_group_id }}</span></p>
+        <p *ngIf="errors?.agentGroup" class="detail-title">{{ errors.agentGroup }}
+          <span class="ns1red">{{ this.dataset.agent_group_id }}</span>
```
manrodrigues, line 29: Is possible display the name instead of the ID? I guess use never have access to the ID on ORB UI

gpazuch: @manrodrigues I don't think so, since the entity has already been deleted from the database, hence the error `unable to retrieve`.
The dataset only keeps reference to the agent group is `id`.
The required information no longer exists.

iatdaitan (approved):

PR #1029, opened 2022-04-04
Fix/1005 orb remove deleted group from dataset db

mclcavalcante:
When deleting an agent group that was linked to a dataset, besides making the dataset invalid, deletes the group from dataset DB


policies/api/http/logging.go
```diff
@@ -284,6 +284,20 @@ func (l loggingMiddleware) DeleteSinkFromAllDatasetsInternal(ctx context.Context
...
+	defer func(begin time.Time) {
+		if err != nil {
+			l.logger.Warn("method call: delete_agroup_from_all_datasets",
+				zap.Error(err),
+				zap.Duration("duration", time.Since(begin)))
+		} else {
+			l.logger.Info("method call: delete_agroup_from_all_datasets",
+				zap.Duration("duration", time.Since(begin)))
+		}
+	}(time.Now())
```
lpegoraro, line 297: this is an awesome way to measure the runtime, I think you could extract that to a function, you will keep the function pure, sending the function name with it, the only thing we would need to worry is the err. Let's talk about it on a separate.

policies/api/http/metrics.go
```diff
@@ -410,6 +410,28 @@ func (m metricsMiddleware) DeleteSinkFromAllDatasetsInternal(ctx context.Context
...
+			"owner_id", ownerID,
+			"policy_id", "",
+			"dataset_id", "",
+		}
+
+		m.counter.With(labels...).Add(1)
+		m.latency.With(labels...).Observe(float64(time.Since(begin).Microseconds()))
+
+	}(time.Now())
+
```
lpegoraro, line 431: awesome metric extraction.

policies/policy_service_test.go
```diff
@@ -1251,6 +1251,79 @@ func TestInactivateDatasetByID(t *testing.T) {
 	}
 }
 
+func TestDeleteAGroupFromDataset(t *testing.T) {
```
lpegoraro, line 1254: we could try to see if we have a stubbing or spy on the logs, to fetch those execution logs

policies/policy_service.go
```diff
@@ -359,5 +359,23 @@ func (s policiesService) InactivateDatasetByIDInternal(ctx context.Context, owne
...
 	}
 
+	return nil
+}
+
+func (s policiesService) DeleteAgentGroupFromAllDatasets(ctx context.Context, aGroupID string, token string) error {
+	ownerID, err := s.identify(token)
+	if err != nil {
+		return err
+	}
```
lpegoraro, line 369: In this layer we usually don't have a log? Is it ensured that someone will log this?

lpegoraro (dismissed):
LGTM

policies/api/http/logging.go
```diff
@@ -284,6 +284,20 @@ func (l loggingMiddleware) DeleteSinkFromAllDatasetsInternal(ctx context.Context
 	return l.svc.DeleteSinkFromAllDatasetsInternal(ctx, sinkID, ownerID)
 }
 
+func (l loggingMiddleware) DeleteAgentGroupFromAllDatasets(ctx context.Context, aGroupID string, token string) (err error) {
+	defer func(begin time.Time) {
+		if err != nil {
+			l.logger.Warn("method call: delete_agroup_from_all_datasets",
```
weyrick, line 290: I like to keep things consistent, and I don't recall us using "agroup" to abbreviate "agent group" before. What do we do in other places? Just "group" or "agent_group"?

policies/api/http/metrics.go
```diff
@@ -410,6 +410,28 @@ func (m metricsMiddleware) DeleteSinkFromAllDatasetsInternal(ctx context.Context
...
 
+func (m metricsMiddleware) DeleteAgentGroupFromAllDatasets(ctx context.Context, aGroupID string, token string) error {
+	ownerID, err := m.identify(token)
+	if err != nil {
+		return err
+	}
+
+	defer func(begin time.Time) {
+		labels := []string{
+			"method", "deleteAGroupFromAllDatasetsInternal",
```
weyrick, line 421: same note on AGroup here. this one should probably match the actual method name anyway

policies/policies.go
```diff
@@ -162,4 +165,7 @@ type Repository interface {
 
 	// DeleteSinkFromAllDatasets removes a sink from a dataset
 	DeleteSinkFromAllDatasets(ctx context.Context, sinkID string, ownerID string) ([]Dataset, error)
+
+	// DeleteAGroupFromAllDatasets removes agent group from a dataset
+	DeleteAGroupFromAllDatasets(ctx context.Context, aGroup string, ownerID string) error
```
weyrick, line 170: again consistency: we shouldn't use "AgentGroup" in one place and "AGroup" in another

policies/policy_service.go
```diff
@@ -359,5 +359,23 @@ func (s policiesService) InactivateDatasetByIDInternal(ctx context.Context, owne
...
+
+	if aGroupID == ""{
```
weyrick, line 371: it looks like you need to run the code formatter

policies/postgres/policies.go
```diff
@@ -567,6 +567,35 @@ func (r policiesRepository) DeleteSinkFromAllDatasets(ctx context.Context, sinkI
 	return datasets, nil
 }
 
+func (r policiesRepository) DeleteAGroupFromAllDatasets(ctx context.Context, aGroup string, ownerID string) error {
+	q := `UPDATE datasets SET agent_group_id = null WHERE mf_owner_id = :mf_owner_id AND agent_group_id = :agent_group_id`
```
weyrick, line 571: This will make the dataset invalid though, so shouldn't we update the status to invalid here too? Or we are counting on other code to do that now?

weyrick (changes requested):

policies/postgres/policies.go
mclcavalcante: > This will make the dataset invalid though, so shouldn't we update the status to invalid here too? Or we are counting on other code to do that now?

Actually, this is already happening, we have only missed this deleting part

weyrick (approved):

PR #1030, opened 2022-04-05
feat(orb-ui): maintenance flag

gpazuch:
* adds MAINTENANCE option for environment variable. Must export it as UNIX timestamp, specified in seconds. This will enable warning that site is under maintenance and block login until timestamp expires.

* adds queryParam `override` to login route. when `override=1`, authorizes user to login, bypassing maintenance warning.

image


iatdaitan:
@weyrick, what do you think about this approach?

weyrick:
Looks ok but what happened if they were logged in previously? Will they get redirected to this automatically?

gpazuch:
> Looks ok but what happened if they were logged in previously? Will they get redirected to this automatically?

No, I don't think so. At least not in the way we would expect it to by orb-ui itself.
This flag is set at build time, only until set expiration date, not necessarily an interval in the future.

I'm not 100% on this, but that level of control would require orb-ui to either be able to check in with some kind of service, or we could just point to a static maintenance page and have devops change routes and log users off 'forcefully', right?

lpegoraro (approved):
LGTM, we will try to use this in the staging environment, to migrate the cluster correctly in prod, we will apply and use the test with a logged in user as well to make sure.

lpegoraro:
just hold off merge when @etaques and @manrodrigues will make the test

PR #1042, opened 2022-04-07
Tests/reset agent remotely

manrodrigues:
- [Remotely reset an agent with policies](integration/reset_agent_remotely.md)
- [Remotely reset an agent without policies](integration/reset_agent_remotely.md)

### DIFF
--- a/python-test/features/steps/control_plane_agents.py
+++ b/python-test/features/steps/control_plane_agents.py
@@ -6,6 +6,7 @@ from behave import given, when, then, step
 from hamcrest import *
 import time
 import requests
+from datetime import datetime
 
 configs = TestConfig.configs()
 agent_name_prefix = "test_agent_name_"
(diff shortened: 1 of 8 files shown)


python-test/features/steps/control_plane_agents.py
```diff
@@ -152,6 +153,15 @@ def check_agent_tags(context, amount_of_tags):
     assert_that(len(dict(agent["orb_tags"])), equal_to(int(amount_of_tags)), "Amount of orb tags failed")
 
 
+@step("remotely restart the agent")
+def reset_agent_remotely(context):
```
lpegoraro, line 157: it would be good to add the Then/When/And even if in a comment above to identify faster when reading

python-test/features/steps/local_agent.py
```diff
@@ -37,6 +37,24 @@ def run_local_agent_container(context, port):
     if port not in context.containers_id.keys():
         context.containers_id[str(port)] = context.container_id
 
+@step('the container logs that were output after {condition} contain the message "{text_to_match}" within'
+      '{time_to_wait} seconds')
```
lpegoraro, line 41: Check if we can use this instead, even though it is just a couple of seconds if we want to run things in parallel in the near future, the sleep can block the execution.
https://realpython.com/python-sleep/#using-eventwait

lpegoraro (dismissed):
Awesome job! Just a few non-blocking comments.

python-test/features/steps/local_agent.py
manrodrigues: @lpegoraro please check if the way i did is the expected way.. If you agree with this way, i'm gonna open a new PR adapting all the tests.. if you have another suggestion or if it is not exactly the way it should be, please let me know

lpegoraro (approved):
LGTM

Issue #546, opened 2021-11-18
Ability to add comma delimited agent tags to orb-agent at startup


iatdaitan:

manrodrigues:
closing based on #1368

PR #1055, opened 2022-04-12
DRAFT - POC To Otel AuthModule

lpegoraro:
This links to #546, we will create a new sinker service exclusively to OTEL, so we will create a POC to prove that concept on a separate scope, to validate the auth then port into ORB.


lpegoraro:
Closing this for now.

PR #1061, opened 2022-04-13
hotfix(orb-ui): #362 force clean cache on every operation

gpazuch:
* force clean cache on every item operation on list-like pages.


lpegoraro (approved):
LGTM

PR #1062, opened 2022-04-13
implement auto backend restart on failure

weyrick:


agent/rpc_from.go
```diff
@@ -135,7 +135,15 @@ func (a *orbAgent) handleDatasetRemoval(rpc fleet.DatasetRemovedRPCPayload) {
...
 func (a *orbAgent) handleAgentReset(payload fleet.AgentResetRPCPayload) {
-	a.Restart(payload.FullReset, payload.Reason)
+	if payload.FullReset {
+		err := a.RestartAll(payload.Reason)
+		if err != nil {
+			a.logger.Error("RestartAll failure", zap.Error(err))
+		}
+	} else {
+		// TODO backend specific restart
+		// a.RestartBackend()
```
lpegoraro, line 145: should we add a warn here?

lpegoraro (approved):
LGTM, I would add a warn in that TODO to remind for a specific restart.

agent/rpc_from.go
weyrick: i think not yet because the RPC doesn't support a specific backend yet, so it can't happen yet.

PR #1063, opened 2022-04-13
Fix/1007 orb remove deleted policy from dataset db

mclcavalcante:
When deleting a policy that was linked to a dataset, besides making the dataset invalid, delete the policy from dataset DB. We are already invalidating the dataset but missed this deleting job


lpegoraro (approved):
LGTM

PR #1098, opened 2022-04-21
feat(orb-ui): few changes to dataset modal

gpazuch:
* name field should be last in form

* policy selector should not appear when creating dataset from policy view.

* group selector should not appear when creating dataset from agent group view(does not exist as of yet)


lpegoraro (approved):
LGTM

PR #1100, opened 2022-04-21
feat(agent): adapt otlphttpexporter to use mqtt

lpegoraro:
Open Telemetry MQTT Exporter and Agent adapted to send otlp metrics to sinker.
Note: the Sinker will not be able to receive yet, this is agent only.

To validate, I did a few regression tests on my own
Setup a regular agent in a deployed orb.

The agent configured with default config and using default scraping, it managed to scrape metrics and display on grafana as shown below.

Screenshot from 2022-05-10 16-16-27


agent/otel/otlpmqttexporter/otlp.go
```diff
@@ -0,0 +1,118 @@
...
+	// Client construction is deferred to start
+	return &exporter{
+		config:    oCfg,
+		logger:    set.Logger,
+		userAgent: userAgent,
+		settings:  set.TelemetrySettings,
+	}, nil
+}
+
+// start actually creates the HTTP client. The client construction is deferred till this point as this
```
weyrick, line 60: minor, fix comment. substitute MQTT for HTTP and describe that client is only created if not passed in?

agent/otel/otlpmqttexporter/otlp.go
```diff
@@ -0,0 +1,118 @@
...
+}
+
+func (e *exporter) pushLogs(_ context.Context, _ plog.Logs) error {
+	return fmt.Errorf("not implemented")
+}
+
+func (e *exporter) export(_ context.Context, metricsTopic string, request []byte) error {
+	// convert metrics to interface
+	body := request
+	if token := e.config.Client.Publish(metricsTopic, 1, false, body); token.Wait() && token.Error() != nil {
```
weyrick, line 111: so I think we need to embed the body in an RPC call, the way we do here
https://github.com/ns1labs/orb/blob/9fef3976bed6c47bb2f733c7931838395f79cb2e/agent/backend/pktvisor/pktvisor.go#L345

agent/config/types.go
```diff
@@ -36,7 +36,9 @@ type Cloud struct {
 }
 
 type Opentelemetry struct {
-	Enable bool `mapstructure:"enable"`
+	Enable       bool       `mapstructure:"enable"`
+	MQTT         MQTTConfig `mapstructure:"mqtt"`
+	MetricsTopic string     `mapstructure:"metrics_topic"`
```
weyrick, line 41: this shouldn't be user configurable from the config file

agent/backend/pktvisor/pktvisor.go
```diff
@@ -62,7 +66,7 @@ func (p *pktvisorBackend) GetStartTime() time.Time {
 
 func (p *pktvisorBackend) SetCommsClient(agentID string, client mqtt.Client, baseTopic string) {
 	p.mqttClient = client
-	p.metricsTopic = fmt.Sprintf("%s/m/%c", baseTopic, agentID[0])
+	p.metricsTopic = fmt.Sprintf("%s/m/%s", baseTopic, agentID)
```
weyrick, line 69: why the switch to full agentID?

weyrick (changes requested):

agent/backend/pktvisor/pktvisor.go
lpegoraro: hm, this was leading to an error, and the message was not sending, not sure how it was before, is this just the first part of the agentID?

agent/config/types.go
lpegoraro: Alright, I was in doubt of that, I used this for local testing.

agent/otel/otlpmqttexporter/otlp.go
lpegoraro: hmm this maybe is the reason sinker is failing.

agent/otel/otlpmqttexporter/otlp.go
lpegoraro: will do!

weyrick (approved):

PR #1127, opened 2022-04-25
wrong text for policies, menu without children, key value as tags

manrodrigues:
Fix wrong text for policy without groups linked, move agents and groups outside of fleet and name key values pair as tag on agents view to maintain the pattern of Orb


gpazuch (approved):
LGTM

Issue #1111, opened 2022-04-22
Policies view: user is allowed to add multiple times the same sink to the dataset


manrodrigues:
user is allowed to add multiple times the same sink to the dataset:

![image](https://user-images.githubusercontent.com/78241475/164606317-de20befa-cc86-4089-8647-53588da8bfdc.png)

manrodrigues:
ok

last commit - https://github.com/ns1labs/orb/commit/c6a30bdf677f30f4bd511fa77c7527fc4db66a0a


![sink_only_once](https://user-images.githubusercontent.com/78241475/166705174-3ddd342b-79c9-4b05-a6d9-cd786e1c51bc.gif)

PR #1132, opened 2022-04-26
fix(orb-ui): #1111 sink control bugs

gpazuch:
* fix unable to remove selected sink

* fix able to select same sink multiple times


ui/src/app/shared/components/orb/sink-display/sink-display.component.ts
```diff
@@ -21,8 +21,9 @@ export class SinkDisplayComponent implements OnInit {
   ngOnInit(): void {
   }
 
-  onRemoveSink(sinkID) {
-    this.sinksChange.emit(this.sinks.filter(sink => sink.id !== sinkID));
+  onRemoveSink(sink) {
+    this.sinks.splice(this.sinks.indexOf(sink), 1);
+    this.sinksChange.emit(this.sinks);
```
manrodrigues, line 26: When a sink is removed, it should be shown again as a choice



remove_sink

manrodrigues (changes requested):
Only a detail about when sinks are handled multiple times on the same action

manrodrigues (approved):
Ok

Issue #1117, opened 2022-04-22
Policies view: name of the button for create a dataset must be  "+ new dataset"


manrodrigues:
Of course this is just  a minor, but for maintain the pattern, the name of the button for create a dataset must be 

"+ new dataset", not "+ create dataset"

![image](https://user-images.githubusercontent.com/78241475/164605302-6f369517-18da-4701-99c4-ca430ac79624.png)

manrodrigues:
ok

![image](https://user-images.githubusercontent.com/78241475/165813690-5e7c0f71-5c8e-42c3-a9db-80442fbe0174.png)

Issue #1116, opened 2022-04-22
Policies view: mouse icon must be a hand icon when over the dataset name, once it is clickable


manrodrigues:

manrodrigues:
ok

![mouse_icon_policy](https://user-images.githubusercontent.com/78241475/165813578-e6112d02-b514-40b8-b082-d562259cc77b.gif)

Issue #1084, opened 2022-04-20
Policies: Mouse icon must change when over policies name


manrodrigues:
When the mouse is over the policy name (once now this is a link to policies view page) the icon should be the hand icon, as we use for agents... look:

![mouse_icon](https://user-images.githubusercontent.com/78241475/163001914-59090e9d-f401-43c6-a39b-04945f9b5e0b.gif)

manrodrigues:
ok

![mouse_icon_policy](https://user-images.githubusercontent.com/78241475/165813803-06ebc77b-5efc-4352-9e03-d20dba87c13a.gif)

PR #1138, opened 2022-04-26
fix(orb-ui): Policy View text button

gpazuch:
* fix(orb-ui): #1117 fix button title '+ new dataset'
* fix(orb-ui): remove letter g from dataset-from component
* fix(orb-ui): #1116 #1084 mouse cursor for elements that are clickable


lpegoraro (approved):
LGTM

PR #1141, opened 2022-04-27
ci(build): split build to speed up build time and enhance step error identification

lpegoraro:
This adds split build and enhances the identification on which step a build error could occur.


ui/src/main.ts
```diff
@@ -1,6 +1,6 @@
 /**
  * @license
- * Copyright Akveo. All Rights Reserved.
+ * Copyright Akveo. All Rights Reserved. test
```
weyrick, line 3: ?

fleet/mocks/agent.go
```diff
@@ -15,20 +15,20 @@ type agentRepositoryMock struct {
 	agentsMock map[string]fleet.Agent
 }
 
-func (a agentRepositoryMock) SetStaleStatus(ctx context.Context, minutes time.Duration) (int64, error) {
+func (a agentRepositoryMock) SetStaleStatus(_ context.Context, _ time.Duration) (int64, error) {
```
weyrick, line 18: was this stuff just to fix warnings?

weyrick (commented):
This is too big of a change for me to comprehend so I will have to trust that the semantics of the build and publish stay the same. We do still need to be able to build just orb-agent when a new pktvisor build is available though, is that still possible?

fleet/mocks/agent.go
lpegoraro: yes and also flag the filter of changes on fleet

ui/src/main.ts
lpegoraro: same, just to trigger the execution on I will remove afterwards

lpegoraro:
> This is too big of a change for me to comprehend so I will have to trust that the semantics of the build and publish stay the same. We do still need to be able to build just orb-agent when a new pktvisor build is available though, is that still possible?

Yeah, basically what we did there was just to split in several jobs. 

The hook of the pktvisor is still in place, this only changes to develop yet. Once this is ok and acknowledged, by all we will move to the other branches

Issue #988, opened 2022-03-25
Policy Name and Description editing on policy view page


iatdaitan:
As a User, I would like to be able to edit the policy Name and Description directly from the policy view page, avoiding me going through all the steps to edit just those specific details.

--

Screenshot: 
![image](https://user-images.githubusercontent.com/86745196/162469479-c8723c15-1707-4021-ac9e-35f3929a8456.png)

lpegoraro:
This can be moved to next sprint if definitions are set with Erika

gpazuch:
latest figma updates
![image](https://user-images.githubusercontent.com/1490938/165595260-27768106-816a-4b31-b525-9cc4d1e1c622.png)

manrodrigues:
ok - last commit - https://github.com/ns1labs/orb/commit/b0b9af0cfe3a60c1ebad6d55f702bcfcc7cfa2b4

![policy_editing_with_view_page](https://user-images.githubusercontent.com/78241475/167181328-912f04fa-f947-4bab-a147-205bac2c7329.gif)

PR #1143, opened 2022-04-27
feat(orb-ui): #988 policy view details inline editing

gpazuch:
inline-policy-details-edit
inline-policy-details-edit-2


gpazuch:
image

manrodrigues (approved):
There is some reason to only discard button is inside the details section?

PR #1145, opened 2022-04-28
ci(build): split build to speed up build time and enhance step error identification

lpegoraro:
This relates to #1141, since it is from a fork, there might be some issues with secrets from GH.


weyrick (approved):

etaques (approved):
LGTM

PR #1170, opened 2022-05-05
Update openapi.yaml

rboucher-ns1:
This is to fix a discrepancy between the API docs and the actual API, specifically the JSON key value for sinks (sink_ids) in the dataset endpoint


gpazuch (approved):
:lgtm: :+1:

PR #1185, opened 2022-05-11
Update fleet openapi.yaml to correct status codes

rboucher-ns1:
Updated the documentation to correctly reflect "200" status codes on GET and PUT (instead of "201")


manrodrigues (approved):
LGTM

PR #1187, opened 2022-05-11
intellij project configs

manrodrigues:
Adding the test project configuration


lpegoraro (approved):
Thanks! Great addition for us to test while working on some bug!

PR #1191, opened 2022-05-13
[FIX] add buildx production pipeline

etaques:
Add buildx on orb-agent to produce images for some architectures when merge on production branch


manrodrigues (approved):
LGTM

Issue #543, opened 2021-11-18
Manage agent taps from control plane/UI


iatdaitan:
Right now these are created only at the agent by config file. We should be able to create/manage from Orb. Requires RPC communication.

gpazuch:
@iatdaitan would this translate as a button or checkbox in agent form, where the user could select bootstrapping the agent via Orb?

weyrick:
This would be part of our expanded functionality on the Agent Detail view. We should be listing the current taps available on that agent, and this ticket would add the functionality that they can create a new Tap as well.

gpazuch:
relates to [#547](https://github.com/ns1labs/orb/issues/547)
blocked until [#520](https://github.com/ns1labs/orb/issues/520) (?)

gpazuch:
I'm a bit unsure on how to estimate this.

iatdaitan:
@gpazuch ignore for now. Let's do a refinement session first.

PR #1192, opened 2022-05-13
Add multiple taps for other input types

leoparente:


etaques (approved):
LGTM.
Once pktvisor support more default taps we could just add them on configuration file leaving them ready to be used, just needs to define a UI strategy to set this configuration.

agent/docker/orb-agent-entry.sh
```diff
@@ -31,6 +31,14 @@ version: "1.0"
 
 visor:
   taps:
+    default_dnstap:
+      input_type: dnstap
+    default_sflow:
+      input_type: flow
```
weyrick, line 37: please add `flow_type` ... defaulting to sflow is only for backwards compatibility which we eventually want to deprecate, we want to be explicit

weyrick (changes requested):

weyrick:
I'm actually not sure how useful this is. If, for example, no IP address to bind to is specified here, you would have to be specified in the policy, which is not useful because it will only apply to one agent. The idea is that these taps abstract host specific details away... so what was the plan to use these?

leoparente:
> I'm actually not sure how useful this is. If, for example, no IP address to bind to is specified here for flow or dnstap, it would have to be specified in the policy, which is not useful because it will only apply to one agent. The idea is that these taps abstract host specific details away... so what was the plan to use these?

The plan is that currently we need to pass a config file when provisioning to be able to run other taps than pcap. So the idea here is to able to start agent then configure the bind/port on policy.

It at least allow do eveything through orb interface without need to use external config file

weyrick:
> The plan is that currently we need to pass a config file when provisioning to be able to run other taps than pcap. So the idea here is to able to start agent then configure the bind/port on policy.
> 
> It at least allow do eveything through orb interface without need to use external config file

hm but that policy will only work on a single agent, right? this seems only useful for testing - it's not a pattern I would want to encourage.

the real solution to this is #543

weyrick:
the only way i can imagine this working is if, just like with pcap which fills in the ehternet interface, we had a way to fill in the bind IP when it ran

leoparente:
Not the best approach to solve our problem. We should create environment variables for bind ip and port

PR #1197, opened 2022-05-17
running behavex

manrodrigues:
- This PR aims to enable scenarios to run in parallel process using [behavex](https://github.com/hrcorval/behavex)
 -- The port conflict is now solved by ~selecting a random free port from 1024 to 65535 (This greatly minimizes the risk of any inconsistency)~ allocating ports considering fork process number (thanks @leoparente )
- It also fixes the problem of listing/removing max 100 elements on which execution

PS: the makefile is not on this PR. I'm going to open another one with it.

Behavex generates an output directory containing a complete html, json and JUnit report. Other than that, an overall_status is released and this can be used for validation of builds on CI process. 

HTML report example:

image

image

Overall file:

image

A great thing is the time:

tests that would be run on 47 minutes are run in about 7 minutes:

image

image


lpegoraro (approved):
LGTM

leoparente (approved):

PR #1198, opened 2022-05-17
fix(agent): fix agent unsubscription and heartbeat after deleting a group

mclcavalcante:
Fix agent unsubscription and heartbeat after deleting a group

- deletes agent policy if it isn't used by other groups matching the agent
- deletes policy and group from agent heartbeat


lpegoraro (approved):
LGTM, well done!

mclcavalcante:
Here are the test results after these changes:
image

weyrick (changes requested):
see https://ns1labs.slack.com/archives/C01UNG6UK37/p1653080602702759

fleet/comms_rpc.go
```diff
@@ -46,6 +46,7 @@ type AgentPolicyRPCPayload struct {
 	Action    string      `json:"action"`
 	ID        string      `json:"id"`
 	DatasetID string      `json:"dataset_id"`
+	GroupID   string      `json:"group_id"`
```
weyrick, line 49: So this needs to be a string[] (array) because during NotifyAgentAllDatasets we will need to fill this out too, and a policy could be coming from multiple groups

fleet/comms_rpc.go
```diff
@@ -60,6 +61,12 @@ type GroupRemovedRPC struct {
 	Payload       GroupRemovedRPCPayload `json:"payload"`
 }
 
+type groupRemovedPolicy struct {
```
weyrick, line 64: Where is this used?

fleet/comms.go
```diff
@@ -88,6 +88,7 @@ func (svc fleetCommsService) NotifyGroupNewDataset(ctx context.Context, ag Agent
 		Version:   p.Version,
 		Data:      pdata,
 		DatasetID: datasetID,
+		GroupID:   ag.ID,
```
weyrick, line 91: This is good but we need to add GroupID to the other places AgentPolicyRPCPayload is created too: NotifyAgentAllDatasets, NotifyGroupPolicyUpdate, NotifyGroupPolicyRemoval and in NotifyAgentAllDatasets there could be multiple GroupIDs

agent/agent.go
```diff
@@ -57,6 +57,7 @@ type orbAgent struct {
 type GroupInfo struct {
 	Name      string
 	ChannelID string
+	policyIDs []string
```
weyrick, line 60: It will be easier to just make this a map keyed by policy ID (with a boolean value of true) to avoid the "contains()" func and the array logic

agent/rpc_from.go
```diff
@@ -126,8 +141,46 @@ func (a *orbAgent) handleAgentStop(payload fleet.AgentStopRPCPayload) {
...
+				if removePolicy, index := contains(groupRemoved.policyIDs, groupPolicy); removePolicy {
+					groupRemoved.policyIDs[index] = groupRemoved.policyIDs[len(groupRemoved.policyIDs)-1]
+					groupRemoved.policyIDs[len(groupRemoved.policyIDs)-1] = ""
+					groupRemoved.policyIDs = groupRemoved.policyIDs[:len(groupRemoved.policyIDs)-1]
+				}
+			}
+		}
+	}
+
+	for _, policy := range groupRemoved.policyIDs {
```
weyrick, line 170: So I think this logic is not right because their policy can "belong" to multiple groups. I think that in addition to storing the list of policy IDs per group, we also need to store the list of group IDs per policy, and only remove the policy if it's not being used by any more groups (otherwise it just removes that group id from the list)

weyrick (changes requested):

fleet/comms_rpc.go
mclcavalcante: got it, but that doesn't mean that we should also fill DatasetID as an array too? since a policy can also be linked to many datasets

lpegoraro (approved):
LGTM

agent/rpc_from.go
```diff
@@ -128,6 +128,25 @@ func (a *orbAgent) handleAgentStop(payload fleet.AgentStopRPCPayload) {
...
+
+		if len(policy.GroupIds) == 0 {
+			err = a.policyManager.RemovePolicy(policy.ID, policy.Name, policy.Backend)
+			if err != nil {
+				a.logger.Warn("failed to remove a policy, ignoring", zap.String("policy_id", policy.ID), zap.String("policy_name", policy.Name), zap.Error(err))
+				continue
+			}
+		}
+	}
+	delete(a.groupsInfos, rpc.AgentGroupID)
```
weyrick, line 148: i think this delete should happen in the `unsubscribeGroupChannel` method which is called above

agent/rpc_from.go
```diff
@@ -128,6 +128,25 @@ func (a *orbAgent) handleAgentStop(payload fleet.AgentStopRPCPayload) {
...
+	policies, err := a.policyManager.GetRepo().GetAll()
+	if err != nil {
+		return
+	}
+
+	for _, policy := range policies {
+		delete(policy.GroupIds, rpc.AgentGroupID)
+
+		if len(policy.GroupIds) == 0 {
+			err = a.policyManager.RemovePolicy(policy.ID, policy.Name, policy.Backend)
```
weyrick, line 141: let's put a log message here: `a.logger.Info("policy no longer used by any group, removing", ...)`

weyrick (changes requested):
almost there, few more minor items! :)

PR #1202, opened 2022-05-18
fix(policy): fix reapply policy after the dataset becomes valid again

mclcavalcante:
fix reapplying policy when the dataset becomes valid again after the sink which was linked is removed and a new one is added

### DIFF
--- a/fleet/comms.go
+++ b/fleet/comms.go
@@ -49,6 +49,8 @@ type AgentCommsService interface {
 	NotifyGroupPolicyUpdate(ctx context.Context, ag AgentGroup, policyID string, ownerID string) error
 	//NotifyAgentReset RPC core -> Agent: Notify Agent to reset the backend
 	NotifyAgentReset(agent Agent, fullReset bool, reason string) error
+	// NotifyGroupDatasetReactivate RPC core -> Agent: Notify Agent an already created Dataset goes active after a while as invalid
+	NotifyGroupDatasetEdit(ctx context.Context, ag AgentGroup, datasetID string, policyID string, ownerID string) error
 }
 
 var _ AgentCommsService = (*fleetCommsService)(nil)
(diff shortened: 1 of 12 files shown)


lpegoraro (dismissed):
LGTM

weyrick (changes requested):
see thread https://ns1labs.slack.com/archives/C01UNG6UK37/p1653071246986939

lpegoraro (approved):
LGTM

weyrick (changes requested):

policies/redis/producer/events.go
```diff
@@ -53,6 +54,15 @@ type removeDatasetEvent struct {
...
 }
 
+type updateDatasetEvent struct {
+	id           string
+	ownerID      string
+	agentGroupID string
+	datasetID    string
+	policyID     string
+	timestamp    time.Time
+}
```
weyrick, line 64: see my other comment: add `valid`, `turnedValid`, `turnedInvalid`

fleet/comms.go
```diff
@@ -69,6 +71,68 @@ type fleetCommsService struct {
...
+	dataset, err := svc.policyClient.RetrieveDataset(ctx, &pb.DatasetByIDReq{
+		DatasetID: datasetID,
+		OwnerID:   ag.MFOwnerID,
+	})
+	if err != nil {
+		return err
+	}
+
+	var action string
+	if dataset.Valid {
```
weyrick, line 94: change to TurnedValid

fleet/comms.go
```diff
@@ -69,6 +71,68 @@ type fleetCommsService struct {
...
+	if err != nil {
+		return err
+	}
+
+	var pdata interface{}
+	if err := json.Unmarshal(p.Data, &pdata); err != nil {
+		return err
+	}
+
+	dataset, err := svc.policyClient.RetrieveDataset(ctx, &pb.DatasetByIDReq{
```
weyrick, line 85: we shouldn't need to call this, it should get all information it needs from the redis stream

weyrick (changes requested):

fleet/redis/consumer/streams.go
```diff
@@ -147,6 +152,32 @@ func (es eventStore) handleDatasetRemove(ctx context.Context, e removeDatasetEve
...
+		datasetID:     read(event, "dataset_id", ""),
+		policyID:      read(event, "policy_id", ""),
+		valid:         readBool(event, "valid", false),
+		turnedValid:   readBool(event, "turned_valid", false),
+		turnedInvalid: readBool(event, "turned_invalid", false),
+	}
+}
+
+func (es eventStore) handleDatasetUpdate(ctx context.Context, e updateDatasetEvent) error {
+	ag, err := es.fleetService.ViewAgentGroupByIDInternal(ctx, e.agentGroupID, e.ownerID)
```
weyrick, line 169: this should be done in the block below so that we avoid a database lookup unless we really need it

weyrick (changes requested):
almost there, one minor optimization and i think we are ready

weyrick (approved):

PR #1220, opened 2022-05-20
Fix/editing policy request validate

mclcavalcante:
Make policy editing more consistent and allow editing policy name and/or description without also sending agent policy or policy data and format


weyrick (approved):

manrodrigues (approved):
Ok

PR #1222, opened 2022-05-20
[FIX] agent build image

etaques:
This PR remove condition pro build agent docker images. We always need to build when we have changes on branch to guarantee new agent version


manrodrigues (approved):
LGTM

PR #1234, opened 2022-05-23
fix(orb-ui): rollback dataset-from rule for showing agent_group_id

gpazuch:
rollback dataset-from rule for showing agent_group_id selector dropdown.
image


manrodrigues (approved):
LGTM

Issue #1216, opened 2022-05-20
Shows unique agent and orb tags


iatdaitan:
As a user, I would like to see unique tags for agents on the dashboard, allowing me to have a clear view of the unique existing tags of each agent.

---

Screenshot:
![image](https://user-images.githubusercontent.com/86745196/169540398-d24dcf18-8fa9-4855-a816-8dd7712587fc.png)

FR001: If the same tag exists on agent_tags and orb_tags, UI should only show it once in every view.

manrodrigues:
![image](https://user-images.githubusercontent.com/78241475/170171624-77cc3302-14fd-4a34-a7ca-84c107f0ab1a.png)

![image](https://user-images.githubusercontent.com/78241475/170171652-8f8d33a5-d43b-46ad-92b0-3700d528a655.png)

Agent tags are no longer being displayed on agent's view page. We need to visualize all the tags on agent's view page.

manrodrigues:
Another thing is:
the tag is filtering by the key.. 

Agent tags are not editable, so if:

1) there is an agent with **agent tag**-> region:us

2) there is a group with region:us tag

3) this agent is edited on UI and an **orb tag** is inserted as-> region:br

The UI will filter the orb one, but the agent will be unsubscribed from the group. 

It sounds as a bug, because the user will think that the orb tag was not created.

manrodrigues:
ok

PR #1235, opened 2022-05-23
fix(orb-ui): #1216 agent view/list show unique tags

gpazuch:
fix(orb-ui): #1216 agent view/list show unique tags (orb_tags, agent_tags).


manrodrigues:
image

image

Agent tags are no longer being displayed on agent's view page. We need to visualize all the tags on agent's view page.

manrodrigues:
Another thing is:
the tag is filtering by the key.. 

Agent tags are not editable, so if:

1) there is an agent with **agent tag**-> region:us

2) there is a group with region:us tag

3) this agent is edited on UI and an **orb tag** is inserted as-> region:br

The UI will filter the orb one, but the agent will be unsubscribed from the group. 

It sounds as a bug, because the user will think that the orb tag was not created.

PR #1245, opened 2022-05-24
Revert otel commit

manrodrigues:


etaques (approved):
LGTM to revert problem with agent to show taps

Issue #1217, opened 2022-05-20
Sink select box field should fit the whole dataset creation's modal box width


iatdaitan:
As a user, I would like to check the whole name of a sinker when selecting it on the Dataset Creation's modal box, making it easier to read the sink's name without word-breaking it.

---

Screenshot:
![image](https://user-images.githubusercontent.com/86745196/169542288-5b3ce623-951f-47da-a02e-e1bee360ffa2.png)

FR001: Sink select box plus "add button" should match the width of the content on the dataset creation's modal box.

gpazuch:
![image](https://user-images.githubusercontent.com/1490938/170595515-6cf7096c-40cb-48cc-b1a8-d13be80360d5.png)
![image](https://user-images.githubusercontent.com/1490938/170595594-7d1ca6ad-fcc6-4f63-95b1-30ecc50cca1e.png)
![image](https://user-images.githubusercontent.com/1490938/170595619-1597bad6-13cc-405e-adbd-d9141e789ae3.png)

manrodrigues:
Look good :) 

![entire_sink_name](https://user-images.githubusercontent.com/78241475/170750015-9823ba9f-a16e-4bbd-a3b5-7d948173bed6.gif)

PR #1265, opened 2022-05-26
fix(orb-ui): #1217 sink selector size

gpazuch:
image

* fix selector size
* fix disable selector when no available sink to choose from


manrodrigues (approved):
LGTM

PR #1273, opened 2022-05-30
Tests/get output

manrodrigues:
Printing tests output


etaques (approved):
LGTM. 
but will be good if have a way to pretty show output json on git actions screen

PR #1274, opened 2022-05-30
feature(sinker): add flow metrics to sinker parser types with unit tests

mclcavalcante:
add parser for flow protocols metrics to scrape as Prometheus format


lpegoraro (approved):
LGTM, in the description of the PR is always a good idea to have an example of the flow as text to help double-checking the mapping

weyrick (approved):
LGTM too if we've verified the metric names are the same as from pktvisor 👍

PR #1277, opened 2022-06-01
improving errors return

manrodrigues:
When a test scenario fails, the assertion message need to be more helpful to identify the problem.

### DIFF
--- a/python-test/features/steps/control_plane_agents.py
+++ b/python-test/features/steps/control_plane_agents.py
@@ -1,7 +1,7 @@
 from test_config import TestConfig
 from utils import random_string, filter_list_by_parameter_start_with, generate_random_string_with_predefined_prefix,\
     create_tags_set, find_files, threading_wait_until, return_port_to_run_docker_container, validate_json
-from local_agent import run_local_agent_container, run_agent_config_file
+from local_agent import run_local_agent_container, run_agent_config_file, get_orb_agent_logs
 from control_plane_agent_groups import return_matching_groups, tags_to_match_k_groups
 from behave import given, then, step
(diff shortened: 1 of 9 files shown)


python-test/features/steps/control_plane_agents.py
```diff
@@ -231,7 +239,7 @@ def reset_agent_remotely(context):
 
 
 @threading_wait_until
-def expect_container_status(token, agent_id, status, event=None):
+def expect_agent_status(token, agent_id, status, event=None):
```
lpegoraro, line 242: I would add `wait` to the name of the function to distinguish in the name that is a longer duration func.

python-test/features/steps/local_agent.py
```diff
@@ -84,6 +89,7 @@ def run_container_using_ui_command(context, status_port):
     rename_container(context.container_id, LOCAL_AGENT_CONTAINER_NAME + random_string(5))
     if context.container_id not in context.containers_id.keys():
         context.containers_id[context.container_id] = str(context.port)
+    threading.Event().wait(10)
```
lpegoraro, line 92: Isnt here the `10` to be the seconds from parameter?

lpegoraro (approved):
Just two comments, but everything else seems good.

python-test/features/steps/local_agent.py
manrodrigues: ok.

python-test/features/steps/control_plane_agents.py
manrodrigues: check again please

leoparente (approved):

Issue #869, opened 2022-02-11
Register: wrong message for invalid regex on full name field


manrodrigues:
when using numbers, accents or characters such as ç, message displayed is `Full name should contain two space separated names` even when the 2 names are inserted... the alert message for wrong regex must be helpful

![image](https://user-images.githubusercontent.com/78241475/153616533-dab7bc5d-37f8-4d8f-b5ab-6f0cb34e1bb7.png)

![image](https://user-images.githubusercontent.com/78241475/153616797-85093a5c-a5b0-4150-a391-53be71046612.png)

![image](https://user-images.githubusercontent.com/78241475/153617008-5938c664-8306-4872-96ca-8988a0c6a9cc.png)

manrodrigues:
@gpazuch  I don't think we should have this "full name should contain" twice.. 

![image](https://user-images.githubusercontent.com/78241475/171854276-a284f45d-7a38-43b3-bd79-5ea7508fc6b3.png)

Issue #778, opened 2022-01-20
Register: token fails and register button doesn't re-enable


manrodrigues:
@gpazuch  Two comments about things that I noticed.
1) one that I can see in your gif as well, that is, even when user can register an account, a token error is displayed. If token is alright, this error must not appear, but if there is a problem with token, user must not be enabled to register an account.
<br>

![image](https://user-images.githubusercontent.com/78241475/148985193-1e9b302f-347d-43b0-b484-034371a1b11a.png)

<br>

![register_account2](https://user-images.githubusercontent.com/78241475/148985901-59e8973a-eb36-4280-a418-92dbb80a774a.gif)

2)if user forget to click on acceptance term, help message is correctly displayed, but user is not enable do register account anymore, because register button keeps disabled

<br>

![register_account](https://user-images.githubusercontent.com/78241475/148986148-1464d1b5-6b2a-4f85-8d10-8098939500e4.gif)

manrodrigues:
i'm just moving the issues to this card, once now we are able to register an account, so, i'm closing #737

gpazuch:
@iatdaitan p1 maybe?

iatdaitan:
@gpazuch For now, guess it is a p2. Let's deal with it after some important issues and polishments.

manrodrigues:
@gpazuch I guess it's ok... i'd like that the buttom only enables when the terms are marked.. but.. not seems to be a problem this way.

![buttom_register_enabled](https://user-images.githubusercontent.com/78241475/171855287-69e628b8-2c20-4eba-b9e0-51d1502ba914.gif)

gpazuch:
@manrodrigues oops

Issue #768, opened 2022-01-14
Register: UI must treat conflict error when user try to register an email already registered on orb


manrodrigues:
API returns correct error (conflict) when user try to register an already registered email, but UI have no span or soft way to fail on these situations. It's necessary to display some mensagem like: "this email is already registered"

[Edited]: I'm inserting this new information:
- Even when email is already registered, if I try to register again, the ironclad email is sent confirming registration

gpazuch:
related to issue #778

manrodrigues:
@gpazuch  Any error message is being displayed:

![not_working_register](https://user-images.githubusercontent.com/78241475/171855144-b9aa96e1-da34-433b-ba9d-860504611eb8.gif)

gpazuch:
https://github.com/ns1labs/orb/pull/1378

unable to connect PR due to existing links

manrodrigues:
ok

![error_conflict](https://user-images.githubusercontent.com/78241475/176450648-e56b8b2e-93bf-44b2-9da0-3581861db918.gif)

PR #1281, opened 2022-06-02
fix(orb-ui): register page fixes

gpazuch:
fix(orb-ui): #778 token fails and register button does not reenable

fix(orb-ui): #869 fix message for invalid name on register page
* also change regex validator to support special characters and \'. .

fix(orb-ui): #768 email already taken register page fix


manrodrigues (approved):
Ok

Issue #1289, opened 2022-06-02
'hyperlink' all relevant information to respective detail view.


gpazuch:
Make it easier to see in detail all related properties of an entity, redirecting to correspondent View page.

i.e.:
User is in Policy View.
Make it possible to click and go to View Page of each Agent Group, Dataset, Policy and Sink that is listed in the page.

manrodrigues:
This looks very useful:

![links](https://user-images.githubusercontent.com/78241475/173118170-329d9d5f-2368-42ac-a1d0-983956290425.gif)

manrodrigues:
- By closing the dataset modal, the agent view's page is reloaded and all boxes expanded are collapsed. It would be better if the modal close without more changes to the page.

manrodrigues:
- The agent matching the group to which the policy is applied could also be clickable:

![agent_link](https://user-images.githubusercontent.com/78241475/173120747-9ed3dfdc-232f-429a-b654-17e0268f8cb8.gif)

PR #1315, opened 2022-06-07
feat(orb-ui): #1289 hyperlink everything.

gpazuch:


lpegoraro (approved):
LGTM, nice work, I was thinking if we could have a more generic component for that, but I haven't found anything like that.

gbottega (approved):

PR #1318, opened 2022-06-08
Feature/orb 1300 add policy status agent list

mclcavalcante:
Adds the following map structure field to response of agent list endpoint, in which the key is the policy ID and the value is the policy information. All retrieve from lastHBData on DB

``
   "policy_status": {
      "43230fb0-9b28-d73c-0d40-efc8446bcfe8":{
          "name": "test",
          "status": "running"
      }
   }
``
image

### DIFF
--- a/fleet/agent_service.go
+++ b/fleet/agent_service.go
@@ -275,15 +275,17 @@ func (svc fleetService) ViewAgentInfoByChannelIDInternal(ctx context.Context, ch
 	return res, nil
 }
 
-func (svc fleetService) GetPoliciesState(ctx context.Context, agent Agent) (map[string]interface{}, error) {
+func (svc fleetService) GetPolicyState(ctx context.Context, agent Agent) (map[string]interface{}, error) {
 
 	jsonHb, err := json.Marshal(agent.LastHBData)
 	if err != nil {
 		svc.logger.Error("failed to marshal heartbeat data", zap.Error(err))
+		return nil, err
 	}
 	var hb Heartbeat
 	if err = json.Unmarshal(jsonHb, &hb); err != nil {
 		svc.logger.Error("failed to unmarshal heartbeat data", zap.Error(err))
+		return nil, err
 	}
 
 	policyState := make(map[string]interface{})
--- a/fleet/agent_service.go
+++ b/fleet/agent_service.go
@@ -10,6 +10,7 @@ package fleet
 
 import (
 	"context"
+	"encoding/json"
 	"github.com/mainflux/mainflux"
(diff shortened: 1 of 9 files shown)


lpegoraro (approved):
LGTM

lpegoraro:
@mclcavalcante Add a print of it running locally in the Network tab in the Web Console of the Browser or from Postman, that ensures and it is easier on the code reviewer

fleet/api/http/endpoint.go
```diff
@@ -11,6 +11,7 @@ package http
 import (
 	"context"
 	"github.com/go-kit/kit/endpoint"
+	"github.com/mitchellh/mapstructure"
```
weyrick, line 14: why are we using this instead of JSON decode?

weyrick (changes requested):

fleet/api/http/endpoint.go
```diff
@@ -290,16 +291,33 @@ func listAgentsEndpoint(svc fleet.Service) endpoint.Endpoint {
 			},
 			Agents: []agentRes{},
 		}
+
 		for _, ag := range page.Agents {
+			policyStatus := make(map[string]interface{})
+			var hb fleet.Heartbeat
+			err = mapstructure.Decode(ag.LastHBData, &hb)
+			if err != nil {
+				return nil, err
```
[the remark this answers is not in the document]
lpegoraro: We will extract the method into the service that's better either way

fleet/agent_service.go
```diff
@@ -273,3 +274,25 @@ func (svc fleetService) ViewAgentInfoByChannelIDInternal(ctx context.Context, ch
 	}
 	return res, nil
 }
+
+func (svc fleetService) GetPoliciesState(ctx context.Context, agent Agent) (map[string]interface{}, error) {
+
+	jsonHb, err := json.Marshal(agent.LastHBData)
+	if err != nil {
+		svc.logger.Error("failed to marshal heartbeat data", zap.Error(err))
```
weyrick, line 282: need to return here on error so it doesn't continue

fleet/agent_service.go
```diff
@@ -273,3 +274,25 @@ func (svc fleetService) ViewAgentInfoByChannelIDInternal(ctx context.Context, ch
...
+	}
+	var hb Heartbeat
+	if err = json.Unmarshal(jsonHb, &hb); err != nil {
+		svc.logger.Error("failed to unmarshal heartbeat data", zap.Error(err))
```
weyrick, line 286: same here

weyrick (changes requested):

weyrick (approved):

PR #1322, opened 2022-06-09
Add restriction function for sinks unwanted payload information

lpegoraro:
Added tests

image

### DIFF
--- a/sinks/api/http/endpoint.go
+++ b/sinks/api/http/endpoint.go
@@ -13,7 +13,25 @@ import (
 	"github.com/go-kit/kit/endpoint"
 	"github.com/ns1labs/orb/pkg/types"
 	"github.com/ns1labs/orb/sinks"
+	"strings"
 )
+
+var restrictiveKeyPrefixes = []string{"secret", "key", "pass", "token"}
+
+func omitSecretInformation(metadata types.Metadata) (restrictedMetadata types.Metadata) {
+	metadata.RestrictKeys(func(key string) bool {
+		match := false
+		for _, restrictiveKey := range restrictiveKeyPrefixes {
+			if match {
+				return match
+			} else {
+				match = strings.Contains(strings.ToLower(key), restrictiveKey)
+			}
+		}
+		return match
+	})
+	return metadata
+}
 
 func addEndpoint(svc sinks.SinkService) endpoint.Endpoint {
 	return func(ctx context.Context, request interface{}) (interface{}, error) {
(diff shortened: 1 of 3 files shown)


sinks/api/http/endpoint.go
```diff
@@ -13,8 +13,26 @@ import (
 	"github.com/go-kit/kit/endpoint"
 	"github.com/ns1labs/orb/pkg/types"
 	"github.com/ns1labs/orb/sinks"
+	"strings"
 )
 
+var restrictiveKeyPrefixes = []string{"secret", "key", "pass", "token"}
```
weyrick, line 19: Hey this is reasonable but still depends on this string matching. I was thinking we could instead use the backend feature config which is a structure that actually has detailed information on the sink config:
https://github.com/ns1labs/orb/blob/89f48afdf0c4e541ea3ea14efdfa772639d9a038/sinks/backend/prometheus/prometheus.go#L24

Here we have a type for password:
https://github.com/ns1labs/orb/blob/89f48afdf0c4e541ea3ea14efdfa772639d9a038/sinks/backend/prometheus/prometheus.go#L63

So, we should be able to instead call this CreateFeatureConfig() method for each backend, and elide when the ConfigFeature.Type is "password"?

weyrick (changes requested):

sinks/api/http/endpoint.go
lpegoraro: ok, I agree, on it.

sinks/api/http/endpoint.go
```diff
@@ -13,8 +13,27 @@ import (
...
+	"strings"
 )
 
+var restrictiveKeyPrefixes = []string{backend.ConfigFeatureTypePassword}
+
+func omitSecretInformation(metadata types.Metadata) (restrictedMetadata types.Metadata) {
+	metadata.RestrictKeys(func(key string) bool {
+		match := false
+		for _, restrictiveKey := range restrictiveKeyPrefixes {
+			if match {
```
weyrick, line 26: since we are matching on explicit keys now, there was no reason to do contains or lower... just a strait match

Makefile
```diff
@@ -40,7 +40,7 @@ define run_test
 endef
 
 define make_docker
-	$(shell [ -z "$(SERVICE)" ] && SERVICE=$(subst docker_,,$(1)))
+	$(eval SERVICE=$(shell [ -z "$(SERVICE)" ] && echo $(subst docker_,,$(1)) || echo $(SERVICE)))
```
weyrick, line 43: what is the purpose of these changes?

weyrick (changes requested):

Makefile
lpegoraro: This enables the running locally the goal 
`make dockers` and all services will build successfully, before it was erroring out, since it in the pipeline it receives the SERVICE env through the command line

weyrick (approved):

Issue #1287, opened 2022-06-02
Change Input for Agent Group Selector on Dataset Create/Edit


gpazuch:
It is too much work to scroll through hundreds of agents.

Implement input with text suggestion and possibly dropdown as well.

![image](https://user-images.githubusercontent.com/1490938/171745550-46dd35db-f8b1-49f7-9913-1f7546d47e03.png)

manrodrigues:
The filter is working. but if the user chose the group using the list, modal fail to identify the group and save button is not enabled.

![failing_to_recognize_group](https://user-images.githubusercontent.com/78241475/173628170-335c339a-83d7-403d-aee0-8142b5520041.gif)

manrodrigues:
This is very good. Nice job! :)

![group_selector](https://user-images.githubusercontent.com/78241475/176451654-821b6569-ffe6-4106-817f-30758b02a584.gif)

PR #1325, opened 2022-06-10
feat(orb-ui): #1287 Autocomplete Selector

gpazuch:
Changes Dataset-From modal component(for create|edit|delete dataset) is Group Selector with
autocomplete/autosuggestion input + selector combination for retrieving desired agent group id.


manrodrigues:
The filter is working. but if the user chose the group using the list, modal fail to identify the group and save button is not enabled.

failing_to_recognize_group

manrodrigues (changes requested):
please, check the event of select the group using the suggested list

manrodrigues (approved):
LGTM

PR #1340, opened 2022-06-15
fix(agent): add logs in all handle rpc call from fleet

lpegoraro:


weyrick (changes requested):
hold on this for now

PR #1341, opened 2022-06-15
feat(agent): Add the retry mechanism to confirm groupMembership requests and agent policies

lpegoraro:
This retry mechanism creates a ticker each time the request is sent, but the handler of the fleet message never receives the information.
When it does receive the message, it notifies the channel to close the go routine and end the retry.

Added these debug logs and the info for re-request.

```
2022-06-16T01:41:42.376Z	DEBUG	agent/rpc_to.go:58	sending group membership request
2022-06-16T01:41:42.694Z	DEBUG	agent/rpc_to.go:112	sending agent policies request
2022-06-16T01:41:42.850Z	DEBUG	agent/rpc_from.go:176	received group membership, setting success in request
2022-06-16T01:42:12.851Z	INFO	agent/rpc_to.go:144	agent not received any policy from fleet, re-requesting
2022-06-16T01:42:12.851Z	DEBUG	agent/rpc_to.go:112	sending agent policies request
```

And gracefully stopping the container using the default signals (SIGTERM, SIGKILL,..)

```
2022-06-16T21:42:43.388Z	INFO	agent/agent.go:154	stopping agent
2022-06-16T21:42:43.388Z	DEBUG	agent/agent.go:178	closing group membership request ticker and channel
2022-06-16T21:42:43.388Z	DEBUG	agent/agent.go:180	closing policy ticker and channel
2022-06-16T21:42:43.388Z	DEBUG	agent/agent.go:155	dump	{"goroutines": 30, "gocalls": 38}
2022-06-16T21:42:45.317Z	INFO	pktvisor/pktvisor.go:411	pktvisor process stopped	{"pid": 19, "exit_code": 0}
2022-06-16T21:42:45.318Z	DEBUG	agent/logging.go:37	DEBUG mqtt log	{"payload": ["[msgids]  ","cleaned up"]}
2022-06-16T21:42:45.318Z	DEBUG	agent/logging.go:37	DEBUG mqtt log	{"payload": ["[client]  ","disconnected"]}
2022-06-16T21:42:45.318Z	DEBUG	agent/logging.go:37	DEBUG mqtt log	{"payload": ["[store]   ","memorystore closed"]}
2022-06-16T21:42:45.242Z	DEBUG	agent/logging.go:37	DEBUG mqtt log	{"payload": ["[pinger]  ","ping check",1.545312734]}
```

I also added the same log to dump the go routines and calls in the current develop agent, and it was this:
```
2022-06-16T22:06:57.552Z	DEBUG	agent/agent.go:140	stopping agent with number of go routines and go calls	{"goroutines": 30, "gocalls": 36}
```

This relates to #1358

### DIFF
--- a/agent/agent.go
+++ b/agent/agent.go
@@ -61,8 +61,9 @@ type orbAgent struct {
 }
 
 const retryRequestDuration = time.Second
-const retryRequestFixedTime = 30
-const retryDurationIncrPerAttempts = 15
+const retryRequestFixedTime = 5
+const retryDurationIncrPerAttempts = 10
+const retryMaxAttempts = 5
 
 type GroupInfo struct {
 	Name      string
@@ -198,8 +199,6 @@ func (a *orbAgent) restartComms() error {
 	if err := a.startComms(cloudConfig); err != nil {
 		return err
 	}
-
-	a.requestReconnection(a.client, cloudConfig)
 	return nil
 }
 
--- a/agent/agent.go
+++ b/agent/agent.go
@@ -135,6 +135,7 @@ func (a *orbAgent) Start() error {
 	}
 
 	if err := a.startComms(cloudConfig); err != nil {
+		a.logger.Error("could not restart mqtt client")
 		return err
 	}
 
@@ -204,7 +205,9 @@ func (a *orbAgent) restartComms() error {
 		return err
 	}
 	a.closeRequestTickers()
+	a.logger.Debug("restarting mqtt client")
 	if err := a.startComms(cloudConfig); err != nil {
+		a.logger.Error("could not restart mqtt client")
 		return err
 	}
 	return nil
--- a/agent/agent.go
+++ b/agent/agent.go
@@ -151,8 +151,7 @@ func (a *orbAgent) Stop() {
 	a.logger.Info("stopping agent")
 	a.hbTicker.Stop()
 	a.hbDone <- true
-	a.groupRequestSucceeded <- true
-	a.policyRequestSucceeded <- true
+	a.closeRequestTickers()
 	a.sendSingleHeartbeat(time.Now(), fleet.Offline)
 	if token := a.client.Unsubscribe(a.rpcFromCoreTopic); token.Wait() && token.Error() != nil {
 		a.logger.Warn("failed to unsubscribe to RPC channel", zap.Error(token.Error()))
@@ -164,6 +163,13 @@ func (a *orbAgent) Stop() {
 		}
 	}
 	a.client.Disconnect(250)
+}
+
+func (a *orbAgent) closeRequestTickers() {
+	a.groupRequestTicker.Stop()
+	a.groupRequestSucceeded <- true
+	a.policyRequestTicker.Stop()
+	a.policyRequestSucceeded <- true
 }
 
 func (a *orbAgent) RestartBackend(name string, reason string) error {
@@ -197,7 +203,7 @@ func (a *orbAgent) restartComms() error {
 	if err != nil {
 		return err
 	}
-
+	a.closeRequestTickers()
 	if err := a.startComms(cloudConfig); err != nil {
 		return err
 	}
--- a/agent/agent.go
+++ b/agent/agent.go
@@ -142,6 +142,8 @@ func (a *orbAgent) Start() error {
 	a.groupRequestSucceeded = make(chan bool)
 	a.policyRequestSucceeded = make(chan bool)
 	go a.sendHeartbeats()
+	go a.checkGroupMembershipResponse()
+	go a.checkAgentPolicyResponse()
 
 	return nil
 }
--- a/agent/agent.go
+++ b/agent/agent.go
@@ -139,8 +139,6 @@ func (a *orbAgent) Start() error {
 
 	a.hbTicker = time.NewTicker(HeartbeatFreq)
 	a.hbDone = make(chan bool)
-	a.groupRequestTicker = time.NewTicker(retryRequestFixedTime * retryRequestDuration)
-	a.policyRequestTicker = time.NewTicker(retryRequestFixedTime * retryRequestDuration)
 	a.groupRequestSucceeded = make(chan bool)
 	a.policyRequestSucceeded = make(chan bool)
 	go a.sendHeartbeats()
--- a/agent/agent.go
+++ b/agent/agent.go
@@ -48,11 +48,21 @@ type orbAgent struct {
 	heartbeatsTopic   string
 	logTopic          string
 
+	// Retry Mechanism to ensure the Request is received
+	groupRequestTicker     *time.Ticker
+	groupRequestSucceeded  chan bool
+	policyRequestTicker    *time.Ticker
+	policyRequestSucceeded chan bool
+
 	// AgentGroup channels sent from core
 	groupsInfos map[string]GroupInfo
 
 	policyManager manager.PolicyManager
 }
+
+const retryRequestDuration = time.Second
+const retryRequestFixedTime = 5
+const retryDurationIncrPerAttempts = 10
 
 type GroupInfo struct {
 	Name      string
@@ -129,6 +139,8 @@ func (a *orbAgent) Start() error {
 
 	a.hbTicker = time.NewTicker(HeartbeatFreq)
 	a.hbDone = make(chan bool)
+	a.groupRequestSucceeded = make(chan bool)
+	a.policyRequestSucceeded = make(chan bool)
 	go a.sendHeartbeats()
 
 	return nil
--- a/agent/agent.go
+++ b/agent/agent.go
@@ -61,8 +61,8 @@ type orbAgent struct {
 }
 
 const retryRequestDuration = time.Second
-const retryRequestFixedTime = 5
-const retryDurationIncrPerAttempts = 10
+const retryRequestFixedTime = 30
+const retryDurationIncrPerAttempts = 15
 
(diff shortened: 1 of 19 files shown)


agent/agent.go
```diff
@@ -149,6 +167,13 @@ func (a *orbAgent) Stop() {
 		}
 	}
 	a.client.Disconnect(250)
+	defer close(a.hbDone)
+	defer close(a.policyRequestSucceeded)
+	defer close(a.groupRequestSucceeded)
+}
+
+func (a *orbAgent) closeRequestTickers() {
+
```
weyrick, line 176: empty?

agent/agent.go
```diff
@@ -149,6 +167,13 @@ func (a *orbAgent) Stop() {
 		}
 	}
 	a.client.Disconnect(250)
+	defer close(a.hbDone)
```
weyrick, line 170: does the defer have an effect here since the function is ending?

agent/rpc_from.go
```diff
@@ -99,6 +99,8 @@ func (a *orbAgent) handleGroupRPCFromCore(client mqtt.Client, message mqtt.Messa
 			return
 		}
 		a.handleAgentPolicies(r.Payload, r.FullList)
+		a.logger.Debug("received agent policies, setting success in request")
```
weyrick, line 102: this messages a little confusing for me, can we make it "received agent policies, marking success"

go.mod
```diff
@@ -170,6 +170,7 @@ require (
 
 //These libs are used to allow orb extend opentelemetry features
 require (
+	github.com/google/uuid v1.3.0
```
weyrick, line 173: not needed i think?

agent/rpc_from.go
```diff
@@ -147,57 +149,61 @@ func (a *orbAgent) handleAgentReset(payload fleet.AgentResetRPCPayload) {
 }
 
 func (a *orbAgent) handleRPCFromCore(client mqtt.Client, message mqtt.Message) {
+	go func() {
```
weyrick, line 152: why is this wrapped in a go function now? i don't see how concurrency helps here

agent/rpc_to.go
```diff
@@ -71,12 +72,48 @@ func (a *orbAgent) sendGroupMembershipReq() error {
...
+				duration := retryRequestFixedTime + (calls * retryDurationIncrPerAttempts)
+				a.groupRequestTicker.Reset(time.Duration(duration) * retryRequestDuration)
+				err := a.sendGroupMembershipRequest()
+				if err != nil {
+					a.logger.Error("failed to send group membership request", zap.Error(err))
+					return
+				}
+			}
+		}
+		a.logger.Warn(fmt.Sprintf("retried %d times and still got no response from fleet", retryMaxAttempts))
```
weyrick, line 105: let's add in this log message that this is specific to group request (as opposed to policy)

agent/rpc_to.go
```diff
@@ -71,12 +72,48 @@ func (a *orbAgent) sendGroupMembershipReq() error {
...
+	}
+	go func() {
+		defer a.groupRequestTicker.Stop()
+		for calls := 1; calls <= retryMaxAttempts; calls++ {
+			select {
+			case <-a.groupRequestSucceeded:
+				a.groupRequestTicker.Stop()
+				return
+			case _ = <-a.groupRequestTicker.C:
+				a.logger.Info("agent has not received any group membership from fleet, re-requesting")
```
weyrick, line 95: let's display the timeout value in this message, i.e. how long we just waited unsuccessfully

agent/rpc_to.go
```diff
@@ -96,3 +133,30 @@ func (a *orbAgent) sendAgentPoliciesReq() error {
...
+				duration := retryRequestFixedTime + (calls * retryDurationIncrPerAttempts)
+				a.policyRequestTicker.Reset(time.Duration(duration) * retryRequestDuration)
+				err := a.sendAgentPoliciesRequest()
+				if err != nil {
+					a.logger.Error("failed to send agent policies request", zap.Error(err))
+					return
+				}
+			}
+		}
+		a.logger.Warn(fmt.Sprintf("retried %d times and still got no response from fleet", retryMaxAttempts))
```
weyrick, line 159: again let's mentioned that this is specific to policies

weyrick (changes requested):
let's not add concurrency where it isn't needed - we're not trying to optimize for speed, and adding concurrency always adds extra opportunity for problems.

agent/agent.go
lpegoraro: Yeah, I went by the book on closing channels with any concurrent issue, this actually ensure that the channel doesn't stay opened.

go.mod
lpegoraro: Yeah, I added because I added some span IDs to track each concurrent request, will remove it.

agent/rpc_from.go
lpegoraro: This is actually a good practice from paho's documentation, when we use the option.OrderMatters(true) which is also the default, it says that that the message handling needs to be a go function. With the tests I did, this closes the connection. 
Here
```go
/ SetOrderMatters will set the message routing to guarantee order within
// each QoS level. By default, this value is true. If set to false (recommended),
// this flag indicates that messages can be delivered asynchronously
// from the client to the application and possibly arrive out of order.
// Specifically, the message handler is called in its own go routine.
// Note that setting this to true does not guarantee in-order delivery
// (this is subject to broker settings like "max_inflight_messages=1" in mosquitto)
// and if true then handlers must not block.
```

In the last line of this it says "and if true then handlers must not block.

https://github.com/eclipse/paho.mqtt.golang/blob/bcb58ee4a1782dd29e40cc37230a1f8a31df1376/options.go#L229-L240

agent/rpc_to.go
lpegoraro: actually, I am looking at the result of the log, and we have the method caller there, don't you think it would be redundant?

agent/agent.go
lpegoraro: I was using for testing, forgot to remove that.

agent/rpc_from.go
weyrick: Hm ok wouldn't it be better to wrap it in a go function here, rather than change handleRPCFromCore:
https://github.com/ns1labs/orb/blob/1bdabbd43c05c6e3f93a75ea32ab694480d318ec/agent/comms.go#L47

agent/rpc_to.go
weyrick: hm i'd need to see an example, do you mean the call stack? i think I would rather have it directly in the message though, i think the method may not show up in some contexts

agent/rpc_from.go
lpegoraro: hm I am used to that way, on the function is better because it can be tracked better, and it allow us to control things like context better if we ever add context to those calls.

agent/rpc_to.go
```diff
@@ -71,12 +72,54 @@ func (a *orbAgent) sendGroupMembershipReq() error {
...
+				a.groupRequestTicker.Reset(time.Duration(duration) * retryRequestDuration)
+				err := a.sendGroupMembershipRequest()
+				if err != nil {
+					a.logger.Error("failed to send group membership request", zap.Error(err))
+					return
+				}
+				lastT = t
+			}
+		}
+		a.logger.Warn(fmt.Sprintf("retryGroupMembership retried %d times and still got no response from fleet", retryMaxAttempts))
```
weyrick, line 111: what actually happens in this case afterwards? should we stop the agent?

agent/rpc_to.go
```diff
@@ -96,3 +139,35 @@ func (a *orbAgent) sendAgentPoliciesReq() error {
 
 	return nil
 }
+
+func (a orbAgent) retryAgentPolicyResponse() {
```
weyrick, line 143: it's a shame (and a maintenance burden) to reproduce all of this complex logic - is there any way to make this a single function used by both group and policy (and maybe others in the future?). maybe we can flag it as tech debt if nothing else.

agent/rpc_to.go
```diff
@@ -96,3 +139,35 @@ func (a *orbAgent) sendAgentPoliciesReq() error {
...
+			a.logger.Info("execution period of the re-request of retryAgentPolicy", zap.Duration("period", time.Now().Sub(t)))
+		}(time.Now())
+		lastT := time.Now()
+		for calls := 1; calls <= retryMaxAttempts; calls++ {
+			select {
+			case <-a.policyRequestSucceeded:
+				a.policyRequestTicker.Stop()
+				return
+			case t := <-a.policyRequestTicker.C:
+				a.logger.Info("agent not received any policy from fleet, re-requesting", zap.Duration("waiting period", lastT.Sub(t)))
```
weyrick, line 159: "agent did not receive any policies from fleet"

agent/rpc_to.go
lpegoraro: I don't think we should stop the agent on this, maybe another reset, can fix, or if there is another issue with mqtt connection this can actually stop the container on other functions, the re-request should not have this responsability

agent/rpc_to.go
lpegoraro: yeah, I am assembling something to demonstrate, I did this on the previous project that fixed most of the issues, by adding names to all routines, and pass along context, and instead of channels for tracking the info with routine and cancel functions. 
and also to divide this code to more adherent to single responsibility principle... I will open a ticket with this.

weyrick (approved):

PR #1349, opened 2022-06-17
quickfix fullmatch name filter

gpazuch:


manrodrigues (changes requested):
This strategy will solve one problem, but cause others.

PR #1352, opened 2022-06-17
Tests/failing scenarios group

manrodrigues:
- Inserting more scenarios that consider agents subscribing to previous existing groups

- Inserting scenario for check removal of policies that failed to apply


manrodrigues:
wrong code attached

PR #1371, opened 2022-06-23
Fix/1219 delete channels when remove group

mclcavalcante:
Also delete MQTT channel when agent group is deleted in orb


lpegoraro (dismissed):
LGTM

weyrick (changes requested):
so this is good but i think it's not enough: we also have to do the opposite of this: https://github.com/ns1labs/orb/blob/7cd17f6ebaba2954083964597f3ae3d7a51e160d/fleet/agent_group_service.go#L56

agents are "connected" to the group channel, and we have to un-connect them. looks like mainflux sdk has a "ThingsByChannel" we can use to get all agents in the group channel, then call "DisconnectThing" on each one

fleet/agent_group_service.go
```diff
@@ -235,6 +235,33 @@ func (svc fleetService) RemoveAgentGroup(ctx context.Context, token, groupId str
...
 
+	group, err := svc.agentGroupRepository.RetrieveByID(ctx, groupId, ownerID)
+	if err != nil {
+		return err
+	}
+
+	limit, err := svc.agentRepo.RetrieveAllByAgentGroupID(ctx, ownerID, groupId, false)
+	if err != nil {
+		return err
+	}
```
lpegoraro, line 246: If you are using this just for the limit, I think we should add a query to fetch just the count, it is way faster.

lpegoraro (dismissed):
LGTM

weyrick (changes requested):

fleet/agents.go
```diff
@@ -139,6 +139,8 @@ type AgentRepository interface {
 	SetStaleStatus(ctx context.Context, minutes time.Duration) (int64, error)
 	// RetrieveAgentInfoByChannelID gRPC version to retrieve ownerID, name and agent tags by a provided channelID
 	RetrieveAgentInfoByChannelID(ctx context.Context, channelID string) (Agent, error)
+	// RetrieveCountGroupAgents retrieve the total agents of an agent group
+	RetrieveCountGroupAgents(ctx context.Context, owner string, tags types.Tags) (int, error)
```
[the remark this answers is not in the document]
mclcavalcante: yes, it was only added for this reason. Ok, I see it, I'm working on this

fleet/agent_group_service.go
```diff
@@ -235,6 +237,37 @@ func (svc fleetService) RemoveAgentGroup(ctx context.Context, token, groupId str
...
+	for {
+		connectedAgents, err := svc.mfsdk.ThingsByChannel(token, group.MFChannelID, lastRead, limitThingsByChannel, false)
+		if err != nil {
+			return err
+		}
+
+		for _, agent := range connectedAgents.Things {
+			err := svc.mfsdk.DisconnectThing(agent.ID, group.MFChannelID, token)
+			if err != nil {
+				return errors.Wrap(errors.New("error while disconnecting agents"), err)
```
weyrick, line 255: i don't think we should return an error here, because it will leave other agents still connected and orphaned. let's log the error but continue through the list

weyrick (changes requested):

fleet/agent_group_service.go
```diff
@@ -235,6 +237,37 @@ func (svc fleetService) RemoveAgentGroup(ctx context.Context, token, groupId str
...
+	for {
+		connectedAgents, err := svc.mfsdk.ThingsByChannel(token, group.MFChannelID, lastRead, limitThingsByChannel, false)
+		if err != nil {
+			return err
+		}
+
+		for _, agent := range connectedAgents.Things {
+			err := svc.mfsdk.DisconnectThing(agent.ID, group.MFChannelID, token)
+			if err != nil {
+				svc.logger.Error("error while disconnecting agent", zap.String("agent_name", agent.Name), zap.String("agent_id", agent.ID), zap.Error(err))
```
weyrick, line 255: we need to log the channel id as well or else we won't be able to investigate from the log message. let's make the message "error while disconnecting agent from channel"

weyrick (changes requested):

PR #1376, opened 2022-06-24
adding new log to policy removal

manrodrigues:
@mclcavalcante this log is only displayed when the group is removed?


mclcavalcante:
> @mclcavalcante this log is only displayed when the group is removed?

yes @manrodrigues , when the group is removed and there isn't any group using this policy anymore

mclcavalcante (approved):

PR #1382, opened 2022-06-24
Sink credentials security improved

lpegoraro:
## Example of how it would look with a 16char key

With Request 

Screenshot from 2022-06-24 15-17-14

Will result in storing password as 

Screenshot from 2022-06-24 15-18-30

Using testing secret.

Secrets using this needs to be 16/32/64 characters long to work.

## WIP Status
- [x] Implementation
- [x] Kubernetes Secrets with orb-helm repo
- [ ] PR for orb-helm 
- [x] Test Adding sinks
- [ ] Test Update sinks
- [ ] Test Sink Credential working with Sinker Service 
- [ ] Test migrations


gbottega (changes requested):

sinks/password_test.go
```diff
@@ -0,0 +1,59 @@
...
+		t.Run(tt.name, func(t *testing.T) {
+			ps := NewPasswordService(logger, tt.key)
+			got, err := ps.EncodePassword(tt.plainText)
+			if err != nil {
+				t.Fatalf("received error on encoding password: %e", err)
+			}
+			t.Logf("storing %s", got)
+			password, err := ps.GetPassword(got)
+			if err != nil {
+				t.Fatalf("received error on encoding password: %e", err)
```
gbottega, line 47: received error on *decoding* password

sinks/password_test.go
```diff
@@ -0,0 +1,59 @@
...
+			}
+			t.Logf("retrieving %s", password)
+			assert.Equalf(t, tt.plainText, password, "Got Decoded Password %s", password)
+			getPassword, err := ps.GetPassword(tt.encodedString)
+			if err != nil {
+				t.Fatalf("received error on encoding password: %e", err)
```
gbottega, line 53: received error on *decoding* password *from encodedString*

Just to differentiate from the other log

migrate/migrate_service.go
```diff
@@ -59,8 +59,28 @@ func (s *serviceMigrate) Up() (err error) {
...
+	index := current
+	for index >= 0 {
+		s.logger.Info(fmt.Sprintf("applying migration %d of %d", index, latest))
+		err = s.migrations[index-1].Down()
+		if err != nil {
+			s.logger.Error(fmt.Sprintf("error on migration down %d of %d", index, latest), zap.Error(err))
+			break
+		}
+		index--
+	}
```
gbottega, line 78: Shouldn't this loop start at the latest version and go down until the current one (or vice versa)? Because currently this command will undo all the migrations.
I am not sure, but at least the logs seem strange to me, because they start at the current version, and are compared to the latest (btw, the latest variable is just used for it), and it would be going backwards like this:
```
applying migration 3 of 5
applying migration 2 of 5
applying migration 1 of 5
applying migration 0 of 5
```

gbottega (changes requested):

migrate/migrate_service.go
fhaze: nope, the migration should always consider the current version and then do something. In this case rolling back all migrations. It is also common to see some migrations tools to support the specific version you want to migrate up or down, but in our scenario the migration svc was coded too simple to support that.

btw the message should be `rolling back migration X of Y` instead of `applying migration X of Y`

gbottega: Right, I will keep this comment to fix the logs.

sinks/password.go
```diff
@@ -0,0 +1,97 @@
...
+	"crypto/rand"
+	"encoding/hex"
+	"go.uber.org/zap"
+	"io"
+)
+
+type PasswordService interface {
+	EncodePassword(plainText string) (string, error)
+	SetKey(newKey string)
+	GetPassword(cipheredText string) (string, error)
```
weyrick, line 16: please name `DecodePassword`

sinks/password.go
```diff
@@ -0,0 +1,97 @@
...
+	nonce, ciphertext := data[:nonceSize], data[nonceSize:]
+	plaintext, err := gcm.Open(nil, nonce, ciphertext, nil)
+	if err != nil {
+		return nil, err
+	}
+	return plaintext, nil
+}
+
+func createHash(key string) string {
+	hasher := md5.New()
```
weyrick, line 94: why md5 here? seems like it should be trivial to use another hash function which returns 32 bytes (to select AES-256) and has less chance of hash collision on the passphrase. CPU time here shouldn't be an issue.

weyrick (changes requested):
looking really good, just a few small things

migrate/migrate_service.go
lpegoraro: I am reworking this to a stack, so we can fill the stack on migrate up and pop of the stack on migrate down.

weyrick (approved):

gbottega (approved):
lgtm

fhaze (approved):

lpegoraro:
@manrodrigues, @etaques  on your queue to deploy this!

wrapped testing, going up and down

fix-migrate

etaques (approved):
LGTM

PR #1384, opened 2022-06-27
fix(orb-ui): css fixes

### LINKED ISSUE #241: Name and tags incompletely displayed
If the name is longer than 9 characters and there are many tags, the text will be displayed incompletely

gpazuch:
#241  
* Column widths and fluidness improved for tables
* Introduced simple horiz. animation on hover to show tags
* Agent Add View 2nd Step(tags) css minor improvements  
* Agent Group Matching agents tables updates.
  

---


ui/src/app/pages/fleet/agents/list/agent.list.component.ts
```diff
@@ -120,25 +120,33 @@ export class AgentListComponent implements OnInit, AfterViewInit, AfterViewCheck
     this.columns = [
```
lpegoraro, line 120: hm I know this would require a bigger change, but I like using this with columns and tables with CSS
it helps specify each data goes where. For future notes
https://www.w3schools.com/css/css_grid_container.asp

gpazuch: :+1: thx, the listing def. needs an overhaul. lets discuss it in more detail.

PR #1397, opened 2022-06-28
feat(agent): enable resumeSub.

lpegoraro:
Added a fallback keep alive of the agent forever if the agent never reconnects, it will keep trying.

Need to test the following scenarios

- [x] Disconnect, wait a few minutes, re-connect internet and see if agent re-connects
- [ ] Disconnect, stop the agent, it should stop gracefully


lpegoraro:
This was fixed in #1443

PR #1416, opened 2022-07-01
same policy applied by two groups

manrodrigues:
- Scenario: Remove one of the datasets that applies the same policy on the agent
- Scenario: Remove one of the groups that applies the same policy on the agent


etaques (dismissed):
LGTM

etaques (approved):
LGTM again

PR #1440, opened 2022-07-07
fix(fleet): remove dataset from policy manager when deleting group, but not removing the policy

mclcavalcante:
In the case of multiples groups applying the same policy and removing one of them also removes its dataset on the agent's heartbeat
image
image

### DIFF
--- a/fleet/comms.go
+++ b/fleet/comms.go
@@ -146,7 +146,7 @@ func (svc fleetCommsService) NotifyGroupNewDataset(ctx context.Context, ag Agent
 		DatasetID:    datasetID,
 		AgentGroupID: ag.ID,
 	}}
-	
+
 	data := AgentPolicyRPC{
 		SchemaVersion: CurrentRPCSchemaVersion,
 		Func:          AgentPolicyRPCFunc,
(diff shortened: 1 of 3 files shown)


lpegoraro (dismissed):
LGTM

fleet/comms_rpc.go
```diff
@@ -62,8 +62,9 @@ type GroupRemovedRPC struct {
 }
 
 type GroupRemovedRPCPayload struct {
-	AgentGroupID string `json:"agent_group_id"`
-	ChannelID    string `json:"channel_id"`
+	AgentGroupID string   `json:"agent_group_id"`
+	ChannelID    string   `json:"channel_id"`
+	DatasetsID   []string `json:"datasets_id"`
```
weyrick, line 67: please check for precedents when creating new fields in a public interface - it looks like we already use "datasets", not "datasets_id" when communicating an array of dataset IDs

https://github.com/ns1labs/orb/blob/328b48b4f71189e0557d2f7a2753b7b9d0fb8dde/fleet/comms_rpc.go#L132

fleet/comms.go
```diff
@@ -332,10 +332,24 @@ func (svc fleetCommsService) NotifyAgentGroupMemberships(a Agent) error {
 }
 
 func (svc fleetCommsService) NotifyGroupRemoval(ag AgentGroup) error {
+	ctx, cancel := context.WithTimeout(context.Background(), time.Second*3)
```
weyrick, line 335: Why 3 seconds on timeout? In any case, if we need this, this should be variable of the service somewhere, not hardcoded here.

weyrick (changes requested):

weyrick (approved):

PR #1446, opened 2022-07-08
Feat/orb 1119 remove dataset when remove policy

mclcavalcante:
Remove all datasets that are linked with a policy when removing the policy

Screenshot from 2022-07-08 15-51-01
After removing the policy:
Screenshot from 2022-07-08 15-51-36


lpegoraro (approved):
LGTM

weyrick (approved):
the functionality looks ok, the inconsistency on the arguments ownerID and policyID is not great. unfortunately it looks like our repo service has become pretty inconsistent though, unfortunately

PR #1449, opened 2022-07-11
[FIX] image tags for prod

etaques:
This PR add tag ${{VERSION}}-${{commithash}} on images from production, currently we had just latest tag.


manrodrigues (approved):
LGTM

PR #1452, opened 2022-07-11
fix(orb-ui): add policy form issues

gpazuch:
* buttons within forms assigned type='button' remove error 'no form connected...'


lpegoraro (approved):
LGTM

PR #1466, opened 2022-07-13
standardizing the wait time

manrodrigues:
standardizing log timeout and status and maintaining container when test fails


lpegoraro (approved):
LGTM

Issue #1456, opened 2022-07-11
Agents view: page is broken for new agents


manrodrigues:
![image.png](https://images.zenhubusercontent.com/61aa14a937730b887dff82b3/35821966-0cf2-42b4-8057-94cfdea88434)

- create an agent
- click on the name to open the agent's view page

manrodrigues:
Good :)

![image](https://user-images.githubusercontent.com/78241475/179057993-cf7fb8dc-22d1-4907-a27f-56f558b03e60.png)

PR #1472, opened 2022-07-14
fix(orb-ui): #1456 Agent View Fix

gpazuch:
* Fix access no undefined property while retrieving backend and backend version from a newly created agent.


manrodrigues (approved):
LGTM

Issue #1457, opened 2022-07-12
Agents view: Reset agent button must only be enabled when the agent is online or stale


manrodrigues:
There is no reason to have the "reset agent" button when the agent is new or offline, because in both cases the reset via API will not be effective since the agent is not running and it may cause confusion. In that cases the button must be disabled. 

![image.png](https://images.zenhubusercontent.com/61aa14a937730b887dff82b3/cb4bde56-3e01-4e4c-8b76-02a0372239ff)

manrodrigues:
ok

![image](https://user-images.githubusercontent.com/78241475/179580594-2982f24d-c806-4250-84e3-43a9f8257771.png)


![image](https://user-images.githubusercontent.com/78241475/179580673-d36debf4-aba5-428c-8bd6-cb5e3dd40e9c.png)


![image](https://user-images.githubusercontent.com/78241475/179581545-bee2fbe1-3c14-40d0-a7d2-927859f26595.png)

![image](https://user-images.githubusercontent.com/78241475/179581635-0c3b4919-b714-49d7-a805-9bc5cd6cd715.png)

PR #1477, opened 2022-07-14
fix(orb-ui): #1457 Enable reset button only when agent is online or stale

brunofariasdeo:
### Description
**Issue:** https://github.com/ns1labs/orb/issues/1457
Enabled reset button only when agent is online or stale

Demo:
image

image


gpazuch (approved):
:lgtm: :+1:

Issue #1015, opened 2022-03-31
Policies: advanced options clickable square is almost invisible 


manrodrigues:
the advanced options expansion square is missing an outline and must have different color (as other clickable squares on the page)

![image.png](https://images.zenhubusercontent.com/61aa14a937730b887dff82b3/0e2195a6-a9ab-4d31-a05a-236710046516)

manrodrigues:
ok

![image](https://user-images.githubusercontent.com/78241475/180459263-68836895-6859-4815-85c5-ab3ae454705a.png)

PR #1490, opened 2022-07-15
fix(orb-ui): #1015 Make advanced options button look clickable

brunofariasdeo:
### Description
**Issue:** https://github.com/ns1labs/orb/issues/1015
In this PR I've made some small adjustements to make advanced options button look clickable. I thought it was a better idea to create a new class so I wouldn't risk to override another nb-accordion-header across the portal.

### Before:
image

### Now:
https://user-images.githubusercontent.com/42921279/179307179-0302f752-b1ec-4b40-a0f1-064cbb3418ed.mp4

### DIFF
--- a/ui/src/app/pages/datasets/policies.agent/add/agent.policy.add.component.html
+++ b/ui/src/app/pages/datasets/policies.agent/add/agent.policy.add.component.html
@@ -198,7 +198,7 @@
                 </div>
                 <nb-accordion>
                   <nb-accordion-item>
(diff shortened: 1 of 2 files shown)


brunofariasdeo:
@gpazuch I thought that putting a border around the whole div would not be a better idea, because it would give an idea to the user that this is a select button, which is not the case.

gpazuch (approved):
nicely done :+1:

PR #1495, opened 2022-07-15
feat(ops) add slack post notification message

etaques:
This PR changes:
- add a git actions to permit post slack notification message in determined channel
- remove suffix from agent "latest" images


manrodrigues (approved):
LGTM

Issue #1471, opened 2022-07-14
All pages: reload button remains selected


manrodrigues:
The button should have a confirmation that it was clicked (a check, for example) and should remove the selection after that

It happens for all the pages:
- agents
- agent groups
- sink
-policies

![button_reload.gif](https://images.zenhubusercontent.com/61aa14a937730b887dff82b3/9925aefb-7193-40c4-84ef-31974c277400
)

manrodrigues:
ok

![reload_page_button](https://user-images.githubusercontent.com/78241475/179582003-3e91fa6a-39c1-48b7-b127-13836b2a9f4c.gif)

PR #1496, opened 2022-07-15
fix(orb-ui): #1471 Remove selection from reload button once clicked and make it rotate while fetching data

brunofariasdeo:
### Description
**Issue**: https://github.com/ns1labs/orb/issues/1471
I have fixed the active/focus effect, styled it a little bit and also added an animation while it is fetching the data.

**Before:**
image

**Now:**

https://user-images.githubusercontent.com/42921279/179325184-91c5582f-5a0f-4336-99ec-d7b45cacd865.mp4


gpazuch (approved):
:+1: 🥇

Issue #1491, opened 2022-07-15
Navigation bar links 


gpazuch:
In navigation bar, links should change colors to highlight the hover action, as well as active route.

- [ ] Change color while hovering
- [ ] Change color for active route

manrodrigues:
ok

![navigation_bar](https://user-images.githubusercontent.com/78241475/179577823-dfec8281-660e-4f9f-aa5d-524a1a0c9a9f.gif)

Issue #1470, opened 2022-07-14
Datasets: dataset modal is not being closed after click out on agent view's page


manrodrigues:
Datasets: dataset modal is not being closed after click out on agent view's page
![dataset_close](https://user-images.githubusercontent.com/78241475/179010008-52861c43-da50-41a5-9188-2864cebf4d48.gif)

_Originally posted by @manrodrigues in https://github.com/ns1labs/orb/issues/1255#issuecomment-1184537436_

PR #1499, opened 2022-07-18
fix(orb-ui): #1491 Change navbar link color while hovering and when route is active

brunofariasdeo:
### Description
**Issue:** https://github.com/ns1labs/orb/issues/1470

- Changed color while hovering
- Change color for active route

**Before:**
image

**After:**
image


gpazuch (approved):
:lgtm: :+1:

Issue #1502, opened 2022-07-18
Policies: Policies are the unique center-aligned elements on orb


manrodrigues:
All other list on orb are left aligned. 

![image.png](https://images.zenhubusercontent.com/61aa14a937730b887dff82b3/bca04f55-9283-4624-9c4e-87f1ac869215
)

Example:
Agents:

![image.png](https://images.zenhubusercontent.com/61aa14a937730b887dff82b3/33f39513-67ba-4f42-bf58-5cd9e997e782)

Groups:

![image.png](https://images.zenhubusercontent.com/61aa14a937730b887dff82b3/98b77dc4-24fe-4352-99b9-60f8e4d14f06)

PR #1505, opened 2022-07-18
fix(orb-ui): #1502 Align policy names at the left of table row

brunofariasdeo:
### Description
**Issue:** https://github.com/ns1labs/orb/issues/1502

Align policy names at the left of table row

**Demo:**
image


gpazuch (approved):
:100:

Issue #1500, opened 2022-07-18
Policies: When a policy fails to be create the user should not be redirect to policy's view page


manrodrigues:
Steps to reproduce:

- Create a policy
- Try to create another policy with the same name

Expected result:
- Error response must be displayed and user must continue on the policy creation page

Current result:
- User is being redirect to a broken policy view page

![policy_conflict.gif](https://images.zenhubusercontent.com/61aa14a937730b887dff82b3/2782e477-aa69-4737-a916-7479f3028327)

PR #1515, opened 2022-07-19
fix(orb-ui): #1500 Handle exception thrown when failing to create or edit a new policy

brunofariasdeo:
**Issue:** https://github.com/ns1labs/orb/issues/1500

### Description
When trying to create a policy which name was already used before, the user was being redirected to a broken policy view page.
I've mapped the service exception that is thrown from Agent Policy component and also removed the notifications from service to component.

**Before:**
image

**After:**

https://user-images.githubusercontent.com/42921279/179792086-5e4fa147-79e6-4a4c-b095-94a9fbd9909b.mp4


gpazuch (approved):
:+1:

PR #1516, opened 2022-07-19
Fix/remove policy stream

mclcavalcante:
Today's behavior deletes the dataset before sending the RPC stream, but dataset info is needed to fill the message and because of that the message was not being published. This result in the agent not removing the policy. Thus this PR extracts the deletion of datasets belonging to specified policy to a separated service and only deleting after fetching the dataset's info


lpegoraro (approved):
LGTM

PR #1530, opened 2022-07-21
fix(orb-ui): fix page issues after filter/poll update

gpazuch:
* fixed expand cycle 'dir' param undefined - root cause of issue before deploy
* fixed condition on expand cycle, adopted scan() in favor of reduce()
* fixed keep watching on observable of another page when not needed.
* fixed dataset add component only querying the first 100 items
* fixed group service, group listing page errors
* fixed sink service
* fixed policy service
* fixed dataset service
* fixed agent policy view dataset list
* improved load time for agent policy view

### DIFF
--- a/ui/src/app/common/services/orb.service.ts
+++ b/ui/src/app/common/services/orb.service.ts
@@ -1,7 +1,5 @@
 import { Injectable, OnDestroy } from '@angular/core';
 import { AgentGroup } from 'app/common/interfaces/orb/agent.group.interface';
-import { Agent } from 'app/common/interfaces/orb/agent.interface';
-import { AgentPolicy } from 'app/common/interfaces/orb/agent.policy.interface';
 import { Dataset } from 'app/common/interfaces/orb/dataset.policy.interface';
 import { Sink } from 'app/common/interfaces/orb/sink.interface';
 import { AgentGroupsService } from 'app/common/services/agents/agent.groups.service';
@@ -13,6 +11,7 @@ import {
   BehaviorSubject,
   defer,
   EMPTY,
+  forkJoin,
   merge,
   Observable,
   Subject,
@@ -21,6 +20,7 @@ import {
 import {
   debounceTime,
   map,
+  mergeMap,
   retry,
   shareReplay,
   switchMap,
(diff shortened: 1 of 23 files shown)


ui/src/app/common/services/orb.service.ts
```diff
@@ -78,7 +63,20 @@ export class OrbService implements OnDestroy {
   }
 
   observe<T>(observable: Observable<T>) {
-    return this.poller$.pipe(
+    const controller = merge(
```
gpazuch, line 66: Reusing this controller instance to all observed things was causing issues.
Now, observed things stop being observed after component unsubscribes its 'async' properties.

ui/src/app/common/services/orb.service.ts
```diff
@@ -103,45 +101,27 @@ export class OrbService implements OnDestroy {
...
-    this.sinks$ = this.observe(
-      this.sink.getAllSinks(),
-    );
+  getAgentListView() {
+    return this.observe(this.agent.getAllAgents());
+  }
 
-    this.agentsTags$ = this.agents$.pipe(
+  getAgentsTags() {
+    return this.observe(this.agent.getAllAgents()).pipe(
```
gpazuch, line 124: once observe() is made into a decorator, I can stop wrapping it around the observed thing and simply add @observe on top of each method.

ui/src/app/common/services/orb.service.ts
```diff
@@ -157,55 +137,80 @@ export class OrbService implements OnDestroy {
...
+    return this.observe(this.group.getAllAgentGroups());
   }
 
   getPolicyListView() {
-    return this.policies$;
+    return this.observe(this.policy.getAllAgentPolicies());
+  }
+
+  getPolicyFullView(id: string) {
+    // retrieve policy
```
gpazuch, line 157: this is the mapping used to assemble the agent policy view.
it was scattered in agent policy view and its child components.

ui/src/app/pages/datasets/add/dataset.add.component.ts
```diff
@@ -187,11 +218,11 @@ export class DatasetAddComponent {
   getAvailableAgentPolicies() {
     return new Promise((resolve) => {
       this.loading[CONFIG.POLICIES] = true;
-      const pageInfo = { ...AgentPoliciesService.getDefaultPagination(), limit: 100 };
```
gpazuch, line 190: this component was only getting the first 100 policies and sinks for user to choose from when creating/editing dataset - fixed

brunofariasdeo (approved):
LGTM
I've tested it locally and I believe that all issues mentioned are gone

ui/src/app/common/services/orb.service.ts
```diff
@@ -157,55 +137,80 @@ export class OrbService implements OnDestroy {
...
+    return this.policy.getAgentPolicyById(id).pipe(
+      mergeMap((policy) =>
+        // need a way to get a dataset linked to a policy without having to filter it out
+        this.dataset.getAllDatasets().pipe(
```
gpazuch, line 161: @lpegoraro would it be feasible to add a filter by 'policy_id' in the dataset list endpoint?

gpazuch:
@brunofariasdeo 

I've pushed fixes to the issue you reported to me about not rendering Agent Policy View page for newly created policies.

While the fixes do solve for this problem, and I've also added extra measure of checking 'agent_group_id' is !== '', it does not fix for one specific case:

* when removing an agent group, it may take a while for dataset to reflect this, causing UI to request a resource that is no longer existent/unretrievable. 

It shouldn't be too hard to add a proper treatment for that, but not without going a bit knee deep on rxjs observables.

I think we are good to merge, and deal with the error in a new 'slab' of code

brunofariasdeo (approved):
looks great, ty!

Issue #1547, opened 2022-07-22
Policies: Unable to add dhcp and net handler


manrodrigues:
Create Policy through wizard does not allow adding handlers other than DNS

![policy_multiple_handlers.gif](https://images.zenhubusercontent.com/61aa14a937730b887dff82b3/8aab2932-d9c7-402f-817f-28f148ca126f)

manrodrigues:
ok

![multiple_handlers](https://user-images.githubusercontent.com/78241475/180797287-8fb3094a-80e4-43b0-87c1-101313ae415e.gif)

PR #1558, opened 2022-07-25
feature(orb-ui): #1547 Only add filter property if it is not empty and if it is valid

brunofariasdeo:
**Issue**: https://github.com/ns1labs/orb/issues/1547

### Description
If a DHCP or NET handler is set, it is created with an undefined filter. My commit ensures to check if is not undefined before checking its length for DNS. 

https://user-images.githubusercontent.com/42921279/180787451-41c273a3-100c-423d-a717-bf75aea4dff8.mp4


gpazuch (approved):
:lgtm:

PR #1563, opened 2022-07-25
feat(orb-ui): #1347 agent view policy column

gpazuch:
image

image

* add column showing aggregate state of agent's policies if any.
* add option to filter agent by policies aggregate state.


brunofariasdeo (approved):
LFG!

Issue #1550, opened 2022-07-22
The agent group name exceeds the field limit


pfbrasi:
**Describe the bug**
The agent group name exceeds the field limit

**To Reproduce**
1. Go to Agent Group > New Agent Group
2. Add a long name in Agent Group Name 
3. Add a description, tag key, and tag value

**Expected behavior**
The field has a character limit to not overlap the content next to it. 

**Screenshots**
![image](https://user-images.githubusercontent.com/83239935/180479983-bb2962c7-351a-4a85-84c8-d7cc44df63b1.png)

**Desktop (please complete the following information):**
 - OS: Windows
 - Browser chrome
 - Version 103

manrodrigues:
ok

PR #1564, opened 2022-07-25
fix(orb-ui): #1550 Display agent group name and description in column direction and set a limit to text container

brunofariasdeo:
**Issue**: https://github.com/ns1labs/orb/issues/1550

### Description
Display agent group name and description in column direction and set a limit to text container

**Before:**

image

**Now:**

image


gpazuch (approved):
:lgtm:

Issue #1122, opened 2022-04-22
Insert alerts on dataset creation page that inform user that a group and a sink is required


manrodrigues:
When there is no group or no sink previously created, the modal of dataset creation must have a message that helps the user to know that this elements are missing, and also we must have links to redirect the user to the group or/and sink page.

manrodrigues:
When no sink exist, the list of sinks is blocked (correctly), but no advertisement informing why it's is blocked (because no sink exist) is displayed:

![no_sink](https://user-images.githubusercontent.com/78241475/179551789-1b00766f-a88c-4b5a-8ffc-ab5f067abb5f.gif)

PR #1569, opened 2022-07-26
fix(orb-ui): #1122 Display a message to user when no agent groups or sinks are available

brunofariasdeo:
**Issue**: https://github.com/ns1labs/orb/issues/1122

### Description
Display a message to user when no agent groups or sinks are available

**Before**
image

**Now**
image


gpazuch (approved):
:lgtm:

Issue #1543, opened 2022-07-22
Policy view empty for existent policies


manrodrigues:
Url: `https://stg.orb.live/`
I'm using policy: `test_policy_name_aBSvfdOkgp`
email: `test_ui@stg.com`

![empty_policy_view_page.gif](https://images.zenhubusercontent.com/61aa14a937730b887dff82b3/ca1c71ea-fc01-4eb6-97dd-e924b0fde4d6)

the problem occurs when one of the groups to which the policy is linked is removed.

manrodrigues:
For groups removed is ok:

![image](https://user-images.githubusercontent.com/78241475/181289076-08f710ac-d720-4ef8-8cca-52257e58d9dd.png)

but i noticed that the page is still empty when the sink is removed:

Check policy ` test_policy_name_aFQEKodwAx ` on test_ui@stg.com account.

![image](https://user-images.githubusercontent.com/78241475/181290362-b7aa04e0-05d0-469d-8cc1-43931b777a75.png)

manrodrigues:
ok

![image](https://user-images.githubusercontent.com/78241475/182206942-a7f10251-9ae6-4ecd-80f0-5baabae2b460.png)

PR #1571, opened 2022-07-26
fix(orb-ui): #1543 Display policies that have a dataset which assigned group was removed

brunofariasdeo:
**Issue**: https://github.com/ns1labs/orb/issues/1543

### Description
Display policies that have a dataset which assigned group was removed

**Before**
image

**Now**
https://user-images.githubusercontent.com/42921279/181127129-e5561cc3-3548-4b03-8789-b9ea9864b743.mp4


brunofariasdeo:
@gpazuch , I found out that the issue was being caused by returning EMPTY. Also, it was creating an array with something like `groups:{0: EMPTY, 1: {groupArray}}`, which was not being handled in some cases.
Let me know which fix you think is most suitable for this issue.

gpazuch (commented):
:+1:

Issue #1573, opened 2022-07-27
Hitting enter should be the same as clicking plus for input box with a plus next to it (filters)


manrodrigues:
Whenever you have an input box with a "plus" next to it, pressing enter should have the same functionality as clicking "plus".

Today what works is tab+enter.. the purpose of this card is to enable that just using enter it works

For all filters:

![image.png](https://images.zenhubusercontent.com/61aa14a937730b887dff82b3/a1818588-1c47-4480-8c2f-f22336f6c6ed)

All tags elements (Agents, groups and sinks):

![image.png](https://images.zenhubusercontent.com/61aa14a937730b887dff82b3/44eadcdf-17bb-4070-a48f-6369148bae13)

manrodrigues:
It's ok for filters

![filters_enter_ok](https://user-images.githubusercontent.com/78241475/181515326-54dfd49e-7813-4714-b713-9a0b04d4c9e3.gif)

But we also need to implement it for tags

manrodrigues:
#1598

PR #1588, opened 2022-07-27
fix(orb-ui): #1573 Listen to enter key event in filter input

brunofariasdeo:
**Issue**: https://github.com/ns1labs/orb/issues/1573

### Description
Make filter input listen to enter key event


gpazuch (approved):
:+1: :100:

PR #1600, opened 2022-07-28
feat(orb-ui): String Search enhancements

gpazuch:
* make substring match case insensitive
* add exact match/substring match toggle for simple string properties and tag property filters.


manrodrigues (dismissed):
LGTM

manrodrigues (approved):
LGTM

PR #1610, opened 2022-07-29
fix(orb-ui): #1543 fix policy view not rendering when deleting all sinks

gpazuch:
* fix policy view not rendering when deleting all sinks it indirectly points to


manrodrigues (approved):
LGTM

Issue #1534, opened 2022-07-21
Agent View Group bubbles


gpazuch:
Make active groups in the Agent View be the same modal as in Agent Policy View.

On Policy View, clicking on group shows matching agents modal

On Agent View, clicking on group shows group details

manrodrigues:
ok

![image](https://user-images.githubusercontent.com/78241475/186906469-9b0b5155-58f4-470d-8963-67b75e90ffdf.png)

![image](https://user-images.githubusercontent.com/78241475/186906614-e7f6c4fc-db5a-4275-a52e-82a3c7708e7c.png)

![image](https://user-images.githubusercontent.com/78241475/186906653-7bc1785c-fcfe-426c-8ba4-1ea1000b83b6.png)

PR #1638, opened 2022-08-09
feat(orb-ui): #1534 agent view groups card

gpazuch:
* make agent view groups look like the one in policy view


lpegoraro (approved):
LGTM

PR #1648, opened 2022-08-12
feature(sinker): enhance sinker subscription to check if that picks s…

lpegoraro:
…peed up.

Add go function to sub each message.


etaques (approved):
LGTM

PR #1664, opened 2022-08-15
fix(agent): unsub from all group channels before restarting comms.

### LINKED ISSUE #1517: Agent Reset should also reset Orb agent and not just backends
When requesting an agent reset, the Orb agent should request a fresh set of policies from the Orb control plane as part of the reset and before resetting and refreshing the backends.

A bug has been observed where a Dataset was deleted but the Orb agent still maintained that information and pushed it to  pktvisor, even after an agent reset (observed on NS1 Orb agent 0.17.0-develop-77c9dd4 with ns1-pktvisor 4.2.0-develop-ec9d0ed).

lpegoraro:
This fixes #1517 

### DIFF
--- a/agent/agent.go
+++ b/agent/agent.go
@@ -199,6 +199,9 @@ func (a *orbAgent) RestartBackend(ctx context.Context, name string, reason strin
 }
 
 func (a *orbAgent) restartComms(ctx context.Context) error {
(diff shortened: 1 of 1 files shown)


manrodrigues (approved):
LGTM

PR #1680, opened 2022-08-18
feat(sinks): Sink Update was not updating redis entry with plain password for sinker to use

lpegoraro:
After update, event was being updated in Redis with hashed password, sinker would not fetch correctly


etaques (approved):
LGTM

etaques (approved):
LGTM

Issue #1602, opened 2022-07-28
UI must not enable users to insert label name with length larger then 64 characters


manrodrigues:
Based on #1601

"Next" button must not be enabled for any element (agent, group, policy, sink, dataset) if the name inserted in creation or editing process is larger then 64 characters. 

A help message must be displayed to help the user to identify the error cause.

manrodrigues:
ok

![less_than_64](https://user-images.githubusercontent.com/78241475/186734062-a8bcfc5a-83c1-4c86-b354-9596bc5d1e97.gif)

PR #1686, opened 2022-08-23
fix(orb-ui): #1602 max length input field

gpazuch:
Add max length validator to name, description and other text inputs.


manrodrigues (approved):
LGTM

PR #1689, opened 2022-08-23
fix(agent): add backoff in waiting pktvisor be in a running state to …

lpegoraro:
…send capabilities.

### DIFF
--- a/agent/backend/pktvisor/pktvisor.go
+++ b/agent/backend/pktvisor/pktvisor.go
@@ -117,10 +117,10 @@ func (p *pktvisorBackend) request(url string, payload interface{}, method string
 		return getErr
 	}
 
-	if res.StatusCode != 200 {
+	if (res.StatusCode < 200) || (res.StatusCode > 299) {
 		body, err := ioutil.ReadAll(res.Body)
 		if err != nil {
-			return errors.New(fmt.Sprintf("non 200 HTTP error code from pktvisord, no or invalid body: %d", res.StatusCode))
+			return errors.New(fmt.Sprintf("non 200-299 HTTP error code from pktvisord, no or invalid body: %d", res.StatusCode))
 		}
 		if len(body) == 0 {
 			return errors.New(fmt.Sprintf("%d empty body", res.StatusCode))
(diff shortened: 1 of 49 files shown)


agent/backend/pktvisor/pktvisor.go
```diff
@@ -412,11 +412,37 @@ func (p *pktvisorBackend) scrapeOpenTelemetry() (err error) {
...
 	}
 
-	err = p.exporter.Start(ctx, nil)
-	if err != nil {
-		p.logger.Error("otel mqtt exporter startup error", zap.Error(err))
-		return err
-	}
+	var errStartExp error
+	go func() {
+		startExpCtx := context.WithValue(ctx, "routine", "startExposrter")
```
lpegoraro, line 417: I believe you will want a cancelFunction here also, this way we will handle things better with cancelling the context after the correct initialization of the Exporters, also there is a typo in `startExposrter`

lpegoraro: And don' t forget to defer the call so the go func returning would cancel the context.

agent/backend/pktvisor/pktvisor.go
```diff
@@ -412,11 +412,37 @@ func (p *pktvisorBackend) scrapeOpenTelemetry() (err error) {
...
+					p.logger.Info("waiting until mqtt client is connected", zap.String("wait time", (time.Duration(i)*time.Second).String()))
+					time.Sleep(time.Duration(i) * time.Second)
+					continue
+				}
+			}
+		}
+		if !ok {
+			p.logger.Error("mqtt did not established a connection, stopping agent")
+			p.Stop(startExpCtx)
+			startExpCtx.Done()
```
lpegoraro, line 442: awesome!
With the cancelFunction, we can call it in this if instead of adding the ctx.Done().

lpegoraro (commented):
@mclcavalcante great stuff, just a few comments.

lpegoraro (commented):
@mclcavalcante  I believe that context thing is enough so we can follow up. Great job!

manrodrigues (dismissed):
LGTM

manrodrigues (approved):
LGTM

PR #1711, opened 2022-08-26
fix(orb-ui): fix unable to save policy

gpazuch:
Fix Agent Policy creation through wizard
* unable to read from undefined form when input configuration already predefined.


manrodrigues (approved):
LGTM

Issue #405, opened 2021-10-26
All pages: Creation of empty tags is enabled


manrodrigues:
It is possible to create tags just with spaces " ":" ".
Tags should have a regex (it's not good practice to allow the user to put whatever they want.)

gpazuch:
I believe this shouldn't happen after PR#550

manrodrigues:
Not working yet

![empty-tags](https://user-images.githubusercontent.com/78241475/143089107-2d60607e-949b-482a-b9c2-ba9e18e85022.gif)

manrodrigues:
Tags should accept 2 behaviors: 1) none tag; 2) valid regex tag;

manrodrigues:
ok

![image](https://user-images.githubusercontent.com/78241475/188898771-19d65e41-0604-45d9-8055-7bbbf8abc586.png)

![image](https://user-images.githubusercontent.com/78241475/188899021-972fe3b7-ea9e-4dbc-8806-80e81388993b.png)


![image](https://user-images.githubusercontent.com/78241475/188898869-d952eb85-e4a7-48f5-83d9-fedab4110bf5.png)

manrodrigues:
Creation of tags with empty values are accepted yet.

We need to guarantee that the value is also not empty, though. I'll open a new issue related. 

![image](https://user-images.githubusercontent.com/78241475/188899775-45124cda-9146-409c-ae6c-ce61f029abe2.png)

![image](https://user-images.githubusercontent.com/78241475/188899854-e82c8582-886a-4380-b347-6a22fd161386.png)

PR #1723, opened 2022-08-30
fix(orb-ui): #405 fix tag input 

gpazuch:
* add validator to tag key input, invalidating the 'add' tag button when tag key is empty or blank spaces.


manrodrigues (approved):
LGTM

Issue #1691, opened 2022-08-24
Matching Agents Dialog Improvements


gpazuch:
* make dialog bigger
* improve column distribution
* make matching agents available to see on 2nd step of agent group wizard.

manrodrigues:
ok

![image](https://user-images.githubusercontent.com/78241475/188890546-2c12f59b-3cd5-4150-aa78-dcae5396de04.png)

PR #1739, opened 2022-09-05
fix(orb-ui): #1691 matching agents table improvements

gpazuch:
* increase table size
* improve column distribution - wider space for name, drop last activity column
* make matching agents table accessible from 2nd step of agent group wizard form


manrodrigues (approved):
LGTM

PR #1745, opened 2022-09-06
fix(agent): add rpc nil safe.

lpegoraro:
Fix sometimes failure on marking success if the cancelFunc has not yet set.


manrodrigues (approved):
LGTM

PR #1764, opened 2022-09-09
Extract geodb and fix container stop

etaques:
This PR extract geodb from geo-db folder and set the trap to monitor SIGINT signal on container entrypoint


manrodrigues (approved):
LGTM

PR #1802, opened 2022-09-23
fix(agent): add nil safe for all agent's chan

mclcavalcante:
add nil safe for channels used on agent to avoid panic


lpegoraro (approved):
LGTM

lpegoraro:
@mclcavalcante check the conflicts so we can merge this!

lpegoraro (approved):
LGTM

PR #1806, opened 2022-09-26
fix(sinks): add tags param to url decode for list sinks endpoint

mclcavalcante:
add missing tag param decode on list sinks endpoint and unit test

### DIFF
--- a/sinks/mocks/sinks.go
+++ b/sinks/mocks/sinks.go
@@ -10,11 +10,10 @@ package mocks
 
 import (
 	"context"
-	"fmt"
 	"github.com/gofrs/uuid"
 	"github.com/ns1labs/orb/pkg/errors"
 	"github.com/ns1labs/orb/sinks"
-	"strings"
+	"reflect"
 	"sync"
 )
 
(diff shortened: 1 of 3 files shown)


weyrick (approved):

sinks/mocks/sinks.go
```diff
@@ -92,12 +91,13 @@ func (s *sinkRepositoryMock) RetrieveAll(ctx context.Context, owner string, pm s
...
 	var sks []sinks.Sink
 
-	prefix := fmt.Sprintf("%s", owner)
 	id := uint64(0)
 	for _, v := range s.sinksMock {
 		id++
-		if strings.HasPrefix(v.MFOwnerID, prefix) && id >= first && id < last {
-			sks = append(sks, v)
+		if v.MFOwnerID == owner && id >= first && id < last {
+			if reflect.DeepEqual(pm.Tags, v.Tags) || pm.Tags == nil {
```
Zach-Johnson, line 98: hmm, does this mean I would need to specify an exact match of all the tags on a sink to filter? Or would I be able to specify just one tag?

weyrick: Actual match is done in postgres here https://github.com/ns1labs/orb/blob/2368b50d54795388e16ba26c0cb425e4d4dc62d6/sinks/postgres/sinks.go#L346

The intention is to specify just one tag - it should work that way since that is how we implement our Agent Groups

Zach-Johnson: Ah, okay, awesome!

mclcavalcante: It does work specifying just one tag, check it out:
image

DeepEqual function was used for API unit test porpoises, I can work on improving these tests in forthcoming sprints

PR #1834, opened 2022-09-28
update policy repo inside agent with policy status after applying on backend

mclcavalcante:
policy repo was being updated before applying on backend so policy state was set as unknown. This way it will show up correctly on front end


manrodrigues (approved):
LGTM

PR #1840, opened 2022-09-30
feat(agent): register backends when configuration is enabled

lpegoraro:
Default is pktvisor, and adapt config file to contain this structure:

```yaml
  backends:
    pktvisor:
      enabled: true
      binary: "/usr/local/sbin/pktvisord"
      # this example assumes the file is saved as agent.yaml. If your file has another name, you must replace it with the proper name
      config_file: "/usr/local/orb/etc/agent.yaml"
    cloudprober:
      enabled: false
      binary: "/usr/local/sbin/cloudprober"

```


lpegoraro:
Close since is out of scope.

Issue #1804, opened 2022-09-26
Policies: After editing a Policy Configuration, all the tags are removed from the policy


manrodrigues:
When a policy configuration is edited through the Orb UI all the policy tags are removed from the policy:

![tags_remove_after_edit.gif](https://images.zenhubusercontent.com/61aa14a937730b887dff82b3/ea1dffdd-b599-4ae6-9d1a-bb7eecd8be88)

manrodrigues:
ok

![policy_tag](https://user-images.githubusercontent.com/78241475/194143477-aa47b571-234c-4daa-9be4-04b8f8317b34.gif)

PR #1845, opened 2022-10-03
fix(orb-ui): #1804 fix tag display on policy view

gpazuch:
* fix tag incorrectly displaying policy tags after editing it.


manrodrigues (approved):
LGTM

PR #1863, opened 2022-10-07
orb-agent improvements on backend error

weyrick:
general improvements to orb-agent's handling of failures in the backend
 * split large pktvisor.go into smaller package files
 * add new "backend state" for tracking backend restarts
 * add more status on backend and policy state to heartbeats
 * don't error out when orb-agent fails to remove a policy from a backend; continue removing from the PolicyManager repo
 * when checking if pktvisor is alive, check both process status AND REST API accessibility
 * if backend is not running, update status for all policies on that backend to Unknown

### DIFF
--- a/agent/agent.go
+++ b/agent/agent.go
@@ -7,6 +7,7 @@ package agent
 import (
 	"context"
 	"errors"
+	"fmt"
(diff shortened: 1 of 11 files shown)


etaques (approved):
LGTM

PR #1888, opened 2022-10-16
feat(maestro) Add maestro service

etaques:
This PR changes:
- added maestro service listening on sinks and sinker redis topics and applying and deleting kubernetes deployments from otel-collector-contrib on demand


lpegoraro (approved):
LGTM

PR #1896, opened 2022-10-19
New flow sinker

leoparente:


weyrick (approved):
did we update types for CLI go code in pktvisor too? i know it's not shown/used in CLI right now but good to keep in sync

leoparente:
> did we update types for CLI go code in pktvisor too? i know it's not shown/used in CLI right now but good to keep in sync

I did not. I will do that

PR #1931, opened 2022-10-31
feat(ui): #1866 agent view backend state

gpazuch:
Add backend section to agent view, where user can explore each backend of the agent.


manrodrigues (approved):
LGTM

PR #1941, opened 2022-10-31
fix(ui): Sink 'backend' property standardization on UI nomenclature.

gpazuch:


gpazuch:
@manrodrigues 

Should we stick to 'type' or fix all instances to reflect sink model and name it 'backend'.

I'll update it as needed.

manrodrigues (approved):

PR #1947, opened 2022-11-02
[Feat] Compress otlp payload with brotli algorithm

etaques:
Using OTLP the mqtt payload is bigger than regular way by some reason.
So this PR changes:
- compressed otlpmqtt payload with "BestCompression" using brotli algorithm, reducing it size in ~8x:

image

Metrics still arriving fine after some hours, no mqtt disconnects anymore:
image


lpegoraro (approved):
LGTM

PR #1952, opened 2022-11-02
[FIX] set to reconnect on OTEL agent

etaques:
This PR changes:
- Set to reconnect on OTEL agent when connection broke, on ConnectionLostHandler
image


lpegoraro (approved):
LGTM

PR #1977, opened 2022-11-04
[Feat] Agent scrap per policy on OpenTelemetry

etaques:
This PR changes:
- Permitting agent scrap open telemetry metrics by policy
- Injecting policyID on metrics attributes to be recovered by otel sinker
- Reduced payload per message once that less data sent per policy this time, instead to send all metrics at once

Tested agent with two policies applied is generating two go routines, one per policy to be scrapped:
image

Sinker working with this agent:
image


agent/backend/pktvisor/pktvisor.go
```diff
@@ -223,7 +223,7 @@ func (p *pktvisorBackend) Start(ctx context.Context, cancelFunc context.CancelFu
 	p.scraper.StartAsync()
 
 	if p.scrapeOtel {
-		p.scrapeOpenTelemetry(ctx)
+
```
weyrick, line 226: This should not be an empty block. It also looks like this PR stops using this config variable.

agent/backend/backend.go
```diff
@@ -57,6 +58,7 @@ type Backend interface {
 	Start(ctx context.Context, cancelFunc context.CancelFunc) error
 	Stop(ctx context.Context) error
 	FullReset(ctx context.Context) error
+	ScrapeOpenTelemetry(ctx context.Context, policyName string) error
```
weyrick, line 61: I don't think it makes sense to for this to be part of the Backend interface. This should happen in the backend during ApplyPolicy, as we're doing now.

agent/backend/pktvisor/scrape.go
```diff
@@ -168,7 +170,8 @@ func (p *pktvisorBackend) scrapeOpenTelemetry(ctx context.Context) {
 						return
 					}
 
-					p.receiver, err = p.createReceiver(ctx, p.exporter, p.logger)
+					metricsPath := strings.Replace(defaultMetricsPath, "__all", policyName, -1)
```
weyrick, line 173: We should build the full path, not replace __all here

agent/otel/otlpmqttexporter/otlp.go
```diff
@@ -99,6 +101,12 @@ func (e *exporter) pushTraces(_ context.Context, _ ptrace.Traces) error {
 	return fmt.Errorf("not implemented")
 }
 
+// extract policy from metrics Request
+func extractPolicy(metricsRequest pmetricotlp.Request) string {
+	p, _ := metricsRequest.Metrics().ResourceMetrics().At(0).ScopeMetrics().At(0).Metrics().At(1).Gauge().DataPoints().At(0).Attributes().Get("policy")
```
weyrick, line 106: It seems this assumes the first metric is a Gauge, I don't think this will work.

weyrick (changes requested):
coverage.out was added and shouldn't be here (right?)

Are the go.mod and go.sum changes necessary?

I think the strategy for implementation needs changes, let's discuss first.

agent/backend/pktvisor/pktvisor.go
etaques: fixed

agent/backend/backend.go
etaques: fixed

agent/backend/pktvisor/scrape.go
etaques: I changed it

agent/otel/otlpmqttexporter/otlp.go
etaques: I fixed looking for gauge metrics for each handler we have

etaques:
> coverage.out was added and shouldn't be here (right?)
> 
> Are the go.mod and go.sum changes necessary?
> 
> I think the strategy for implementation needs changes, let's discuss first.



> coverage.out was added and shouldn't be here (right?)
> 
> Are the go.mod and go.sum changes necessary?
> 
> I think the strategy for implementation needs changes, let's discuss first.

Changes on go.sum and go.mod are added after added brotli lib for compression on payload

lpegoraro:
@weyrick, this way worked, because the other method has some flaws in adding information, the insertAttr and extractAttr that @etaques made is correct, the multiple otel pipelines per policy is ok, we can adapt other things afterwards.

agent/otel/bridgeservice.go
```diff
@@ -0,0 +1,46 @@
...
+type bridgeService struct {
+	policyRepo policies.PolicyRepo
+	OrbTags    string
+	AgentTags  string
+}
+
+func NewBridgeService(policyRepo *policies.PolicyRepo) *bridgeService {
+	return &bridgeService{
+		policyRepo: *policyRepo,
+		OrbTags:    "testing-orb",
```
weyrick, line 30: should be removed?

agent/otel/otlpmqttexporter/otlp.go
```diff
@@ -95,18 +108,69 @@ func (e *exporter) start(_ context.Context, _ component.Host) error {
 	return nil
 }
 
+// extractAttribute extract attribute from metricsRequest metrics
+func (e *exporter) extractAttribute(metricsRequest pmetricotlp.Request, attribute string) string {
+	metrics := metricsRequest.Metrics().ResourceMetrics().At(0).ScopeMetrics().At(0).Metrics()
+	for i := 0; i < metrics.Len(); i++ {
+		metricItem := metrics.At(i)
+		if metricItem.Name() == "dns_wire_packets_tcp" || metricItem.Name() == "packets_ipv4" || metricItem.Name() == "dhcp_wire_packets_ack" || metricItem.Name() == "flow_in_udp_bytes" {
```
weyrick, line 116: hm we can't hard code metric names in here. i guess the problem is we otherwise don't know how to pull out the attribute, because the metric type is dynamic?

sinker/otel/bridgeservice/bridge.go
```diff
@@ -4,6 +4,7 @@ import (
 	"context"
 	"sort"
 	"time"
+  "strings"
```
weyrick, line 7: not used?

weyrick (changes requested):

agent/policyMgr/manager.go
```diff
@@ -95,7 +96,7 @@ func (a *policyManager) ManagePolicy(payload fleet.AgentPolicyRPCPayload) {
 				a.logger.Error("failed to retrieve policy", zap.String("policy_id", payload.ID), zap.Error(err))
 				return
 			}
-			if currentPolicy.Version >= pd.Version && currentPolicy.State == policies.Running {
+			if currentPolicy.Version > pd.Version && currentPolicy.State == policies.Running {
```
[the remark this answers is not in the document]
etaques: This is because we should ensure that collector will be dropped and started again during the FullReset, otherwise collector will lost mqtt connection

agent/otel/otlpmqttexporter/otlp.go
etaques: Initially I was getting from a Gauge metric that have on our handlers. But now once we are injecting attribute on all requestMetrics otlp object, consequently in all metrics type exist, perhaps now we dont need more metric name, can just extract metric attribute from the first metric found.

agent/otel/bridgeservice.go
etaques: This was a preparation test to inject orb tags into requestMetrics otlp object. I belive that the intention is to conclude it tomorrow.

agent/policyMgr/manager.go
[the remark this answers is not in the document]
lpegoraro: I am trying to work around the client disconnection on a separate note.

agent/otel/otlpmqttexporter/otlp.go
lpegoraro: I fixed it, there is a different approach, by using the type of the metric.

sinker/otel/bridgeservice/bridge.go
lpegoraro: that was a conflict resolution, resolved.

agent/otel/bridgeservice.go
lpegoraro: I am leaving this for now, once my addition to pass that information to the agent, is complete, we can switch, better than nothing, at least we can start testing, expecting values there

weyrick (approved):
ok thanks for addressing everything here guys, I'm happy with this version 🚀

manrodrigues (approved):

PR #1979, opened 2022-11-05
Feat/add grpc function on policies

etaques:
This PR added a GRPC function to retrieve all datasets per policy, to be consumed on otel sinker micro service.

Tested on BloomGRPC:
image

Tested otel-sinker consume in k8s cluster, and works properly:
image


lpegoraro:
I understand where you are coming from, but I think this is more viable coming from the Agent, and no GRPC would be needed. Let's talk tomorrow about this.

etaques:
> the Agent

Yeah, I see. I created functions on agent to inject attributes (policy_id) on metricsRequest object before export, and it is scrapping metrics by policy, however with this GRPC call on sinker im able to catch all sinksID and send metrics propperly as you did with agentGroups, but now with right logic. The main problem is on the agent part the  receiver is impossible to stop, during the RemovePolicy() or FullReset(), even I closing the goroutine (using context or channel) it still scrapping freakening, seems that receiver open it own go routine.

policies/policies.go
```diff
@@ -124,6 +124,9 @@ type Service interface {
 
 	// ListDatasetsByGroupIDInternal gRPC version of retrieving list of datasets belonging to specified agent group with no token
 	ListDatasetsByGroupIDInternal(ctx context.Context, groupIDs []string, ownerID string) ([]Dataset, error)
+
+	// ListDatasetsByPolicyName gRPC version of retrieving list pf datasets belonging to specified policy
+	ListDatasetsByPolicyID(ctx context.Context, policyID string, ownerID string) ([]Dataset, error)
```
weyrick, line 129: since this is gRPC only, please name `ListDatasetsByPolicyIDInternal`

policies/api/http/metrics.go
```diff
@@ -39,6 +40,23 @@ func (m metricsMiddleware) ListDatasetsByGroupIDInternal(ctx context.Context, gr
 	return m.svc.ListDatasetsByGroupIDInternal(ctx, groupIDs, ownerID)
 }
 
+func (m metricsMiddleware) ListDatasetsByPolicyID(ctx context.Context, policyID string, ownerID string) ([]policies.Dataset, error) {
+	defer func(begin time.Time) {
+		labels := []string{
+			"method", "listDatasetsByPolicyID",
+			"owner_id", ownerID,
+			"policy_id", "",
```
weyrick, line 48: we can use `policyID` here

weyrick (changes requested):
Just to confirm - we will have the ownerID available when we need to make this RPC call?

policies/policies.go
etaques: Yes we have ownerID because we are extracting agent informations by channel on MessageInbound on [https://github.com/ns1labs/orb/blob/c16256a1ae708513ac10659bffdbb657ef37b300/sinker/otel/orbreceiver/otlp.go#L128](url)

the call to this function will be like this:
`sinkIds, err := r.sinkerService.GetSinkIdsFromPolicyID(execCtx, agentPb.OwnerID, policyID)`

The test that I did on my fork on sinker code (that i not opened PR here yet) is consuming this GRPC function as below:    
    
		    ......
		    // Add tags in Context
		    execCtx, execCancelF := context.WithCancel(r.ctx)
		    defer execCancelF()
		    agentPb, err := r.sinkerService.ExtractAgent(execCtx, msg.Channel)
		    if err != nil {
			    execCancelF()
			    r.cfg.Logger.Info("No data extracting agent information from fleet")
			    return
		    }
		    sinkIds, err := r.sinkerService.GetSinkIdsFromPolicyID(execCtx, agentPb.OwnerID, policyID)
		    if err != nil {
			    execCancelF()
			    r.cfg.Logger.Info("No data extracting sinks information from policies, policy ID=" + policyID)
			    return
		    }
		    ....

lpegoraro:
Let's use this approach, this way won't add more complexity

lpegoraro:
we managed to send datasets from the agent. this wont be necessary, closing until otherwise

PR #1990, opened 2022-11-10
Dispatch deploy to stg when necessary (#1990) deploy to stg when necessary

etaques:
Add deploy to stg when necessary


leoparente (approved):
LGTM

PR #2008, opened 2022-11-15
[FIX] otel sinker panic

etaques:
This PR fixes 
- Otel sinker component panic when we are extracting attribute from metricsRequest object
- Sort datasetIDs before inject on metrics attributes on agent side
- OpenTelemetry log message, that was showing when otel was disabled


lpegoraro (approved):
LGTM

PR #2013, opened 2022-11-16
[FIX] Remove datasetIDs from metricsRequest object

etaques:
This PR removes datasetIDs from metricsRequest (otel) attributes, to be sent to prometheus without dataset_ids attribute on metrics

Metrics attributes before change:
image

Metrics attributes after change:
image

### DIFF
--- a/sinker/otel/orbreceiver/otlp.go
+++ b/sinker/otel/orbreceiver/otlp.go
@@ -237,8 +237,9 @@ func (r *OrbReceiver) MessageInbound(msg messaging.Message) error {
 		}
 		datasetIDs := strings.Split(datasets, ",")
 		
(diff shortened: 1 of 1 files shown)


sinker/otel/orbreceiver/otlp.go
```diff
@@ -205,6 +236,9 @@ func (r *OrbReceiver) MessageInbound(msg messaging.Message) error {
 			return
 		}
 		datasetIDs := strings.Split(datasets, ",")
+		
+		// Delete datasetsIDs from metricsRequest
+		mr = r.deleteAttribute(mr, "dataset_ids")
```
lpegoraro, line 241: add policy_ids here also!

lpegoraro (approved):
LGTM

PR #2028, opened 2022-11-18
fix(sinker): remove agent groups and ownerID from metrics.

lpegoraro:


leoparente (approved):
LGTM

PR #2053, opened 2022-11-28
fix(sinks): quick fix sinks api

mclcavalcante:
Remove the need always to send all attributes when updating a sink
Current sink:
image

Editing only sink name:
image

Editing only description:
image

Editing only tags:
image


lpegoraro (approved):
LGTM

PR #2054, opened 2022-11-28
Run coverage only when have changes on microservices

etaques:
This PR back the test coverage to run only when have changes on orb files


lpegoraro (approved):
LGTM

Issue #1985, opened 2022-11-09
Sink API endpoint should support partial updates


rboucher-me:
The sink API endpoint `/api/vi/sinks/<sink-id>` should support modification of _some_ attributes without having to provide _all_ the attributes. This is needed because the password cannot be retrieved and a complete JSON body cannot be assembled for a pre-existing sink.

Updates to the following fields should be supported (without requiring all the fields to be provided in the JSON body):
- `name`
- `description`
- specific tag instances under `tags`

mclcavalcante:
Current behavior:
- If the tags field is omitted, all tags are deleted
- If the tags field is empty, all tags are deleted
- If we already have the tags {test:true} and send an update request with JSON body with tags:{test:false}. It will be overwritten with the newest
- If we already have the tags {test:true} and send an update request with JSON body with tags:{tag:true}. It will delete the oldest tag and save the newest 

Proposed behavior:
- If the tags field is empty, all tags are deleted
- If the tags field is omitted, nothing changes
- If we already have the tags {test:true} among others and send an update request with JSON body with tags:{test:false}. It will only overwrite the one with the same key with the newest value and keep the others as it was
- If we already have the tags {test:true} among others and send an update request with JSON body with tags:{tag:true}. It will append to the preexistent ones and the result will be tags:{test:true, tag:true}

rboucher-ns1:
@mclcavalcante I agree with your proposed behavior.

manrodrigues:
ok.. 

My only concern is about remove tags..

Thinking about updating new tags, ok.. 
but if let's imagine it:

I already have a policy with these policies:

```
    "tags": {
        "my": "tag",
        "other": "tag",
        "test": "mine"
    }
```

Then I send a PUT request to edit the tags because I noticed that I inserted the wrong tag and I want to change my set of tags:

```
    "tags": {
        "my": "tag",
        "other": "tag"
    }
```

Instead of using only the last both tags, nothing will change.. And thinking in edit it through the ORB UI (remove the bubble) it will look like a bug for the users.
What do you think about it?

and I also have a question:

- why the name is always required on the editing process? In my thoughts, we should enable users to make put requests (to edit) with only name, only description and only tags..

mclcavalcante:
Following up on Amanda's comment, there is a use case she came up with that is important and we've missed. Which is a partial update removing just one existent tag. To achieve this with current business logic it is needed to delete all tags and then update with the right ones, which doesn't seem good and useful.
[...]

PR #2063, opened 2022-11-30
refactor(sinks): add business logic to append more tags or remove all when editing a sink

mclcavalcante:
Changes following these rules to tags management on sink API
https://github.com/ns1labs/orb/issues/1985#issuecomment-1331085427

Initial sink
image

Editing tag value
image

Editing and adding a new tag
image

Editing with tags field empty
image


sinks/sinks_service_test.go
```diff
@@ -118,18 +118,49 @@ func TestUpdateSink(t *testing.T) {
...
+		Error:       "",
+		Config:      map[string]interface{}{"remote_host": "data", "username": "dbuser"},
+		Tags:        map[string]string{"cloud": "aws"},
+	}
+	newVTagSink, err = service.CreateSink(context.Background(), token, newVTagSink)
+	require.Nil(t, err, fmt.Sprintf("unexpected error: %s", err))
+	newVTagSink.Backend = ""
+
+	newVTagSink.Tags = map[string]string{"cloud": "true"}
+	newVTag := types.Tags{"cloud": "true"}
```
lpegoraro, line 159: I think, better naming of variables will be a good thing here.
iex. newTags, newTagsSink, newVTagSink, etc.. 
Try using the use-case name of each, and some you can add directly in the case struct, like.
```golang
                   "update sink tags with new tags": {
			sink:  newTagsSink,
			tags:  types.Tags{"cloud": "aws", "test": "true"}, // Eliminating the need of newTags
			token: token,
			err:   nil,
		},
```

sinks/sinks_service_test.go
```diff
@@ -118,18 +118,49 @@ func TestUpdateSink(t *testing.T) {
...
 
 	cases := map[string]struct {
 		sink  sinks.Sink
+		tags  types.Tags
 		token string
 		err   error
 	}{
```
lpegoraro, line 166: Are you using the tags, just for validation of tags?
If so, I would use something like this:
```suggestion
    cases := map[string]struct {
         incomingSink sinks.Sink
         expectedTags types.Tags
         token string
         err error
         }{
```

sinks/sinks_service_test.go
```diff
@@ -168,15 +200,29 @@ func TestUpdateSink(t *testing.T) {
 			token: token,
 			err:   nil,
 		},
+		"update sink tags with new tags": {
+			sink:  newTagsSink,
+			tags:  newTags,
+			token: token,
+			err:   nil,
+		},
```
lpegoraro, line 208: I understand you are using variables like: newTags for multiple places, but for test cases, let's be more readable friendly instead of optimal, so we don' t need to go back and forth reading the test case.

lpegoraro (commented):
Good stuff! I only think we can improve the readability of the test case.

sinks/sinks_service_test.go
mclcavalcante: yes, I'm using it for that, I'll use your suggestion

weyrick (dismissed):

etaques (approved):
LGTM

PR #2067, opened 2022-12-02
feat(sinker): fix inject/delete attributes removing At(0).

### LINKED ISSUE #2066: OTEL metrics are not being scraped (stg environment)
The problem is not on the agent side, since the same agent is working fine in production.
Agents without OTEL enabled are working fine in both env.
So, the problem is restricted to OTEL metrics in stg.. the sink status is unknown and the metrics are not available in grafana.

lpegoraro:
this attempts to fix #2066


manrodrigues (dismissed):
lgtm

etaques (approved):
LGTM

PR #2078, opened 2022-12-12
using agents.orb.live as mqtt address

manrodrigues:
- Changing the mqtt address to use agents.orb.live
- Adding tests to validate default pktvisor configs in agents without auto provisioning


etaques (approved):
LGTM

PR #2112, opened 2022-12-21
make config field mandatory on sinks creation

mclcavalcante:
Make sink config mandatory on sink creation and check if it has the minimum amount of keys to create

*Plus make some unit tests more readable

### DIFF
--- a/sinks/api/http/requests.go
+++ b/sinks/api/http/requests.go
@@ -46,10 +46,13 @@ func (req addReq) validate() error {
 		}
 		//currently, with only prometheus, 2 keys is enough, maybe change latter
 		if keySize >= 2 {
+			//minimal number of keys passed, valid config
 			return true
 		}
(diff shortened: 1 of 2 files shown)


sinks/api/http/requests.go
```diff
@@ -37,6 +37,22 @@ func (req addReq) validate() error {
...
+		return errors.ErrMalformedEntity
+	} else if !req.Config.IsApplicable(func(key string, value interface{}) bool {
+		if key != "" {
+			keySize++
+		}
+		//currently, with only prometheus, 2 keys is enough, maybe change latter
+		if keySize >= 2 {
+			return true
+		}
+		return false
```
lpegoraro, line 51: not entirely sure what this is doing, can you add some comments?

mclcavalcante: just added it, let me know if is still not clear, I can refactor

lpegoraro (approved):
LGTM

PR #2120, opened 2022-12-26
Fix/remove mandatory name editing

mclcavalcante:
allow partial updates by removing mandatory name for editing requests for agents, agent groups, and datasets. following these rules:

- If the name is informed, the new name must overwrite the old name
- If the name is omitted, the old name must be maintained.
- If the name is informed with empty, the old name must be maintained.
- If the name is informed with an invalid value, an error must be raised informing that the name doesn't match the pattern required.


lpegoraro (approved):
LGTM

PR #2127, opened 2022-12-28
Fix/partial updates group

mclcavalcante:
Allow partial updates on editing requests for groups, following these rules:

**Tags:**
- If the tags field is omitted, nothing changes
- If the tags field is informed and empty, an error must be raised informing that tags are required for groups
- If the tags field is informed and NOT empty, a new set of tags overwrite the old one

**Name:**
- If the name is informed, the new name must overwrite the old name
- If the name is omitted, the old name must be maintained.
- If the name is informed but invalid values, an error is raised telling that the name doesn't match the pattern required.
- If the name informed is empty, an error must be raised telling that the name doesn't match the pattern required.

**Description:**
- If the description field is omitted, nothing changes
- If the description field is informed and empty, nothing changes
- If the description field is informed and NOT empty, the new description must overwrite the old one


lpegoraro (approved):
LGTM

PR #2130, opened 2022-12-29
hotfix(ui): change referrer policy to 'origin'

gpazuch:
workaround for mainflux service appending reset password url + token to referer url


lpegoraro (approved):
LGTM

PR #2137, opened 2023-01-04
metric groups test

manrodrigues:
- validating if policies with each metric groups apply and scrapes successfully


etaques (approved):
LGTM

Issue #2099, opened 2022-12-19
When the tags field is omitted in the editing process, they must not be modified


manrodrigues:
When the tags field is omitted in the editing process, they must not be modified.

For now, if tags are omitted on PUT request to edit agents and policies, the tags are removed.

Expected behavior:

PUT requests to /agents/id and /policies/dataset/id that does not contains "tags" must not remove the already existent tags.

- If the tags field is omitted, nothing changes
- If the tags field is informed and empty, all tags are deleted
- If the tags field is informed and NOT empty, new set of tags must overwritten the old one

_Originally posted by @manrodrigues in https://github.com/ns1labs/orb/issues/2005#issuecomment-1358380465_

PR #2139, opened 2023-01-05
[ORB-223] Fix/partial updates policies

mclcavalcante:
- Make it possible to edit tags without any other field being informed in the body, following #2099

- Description:
  - If the description field is omitted, nothing changes
  - If the description field is informed and empty, The description becomes empty (is removed)
  - If the description field is informed and NOT empty, the new description must overwrite the old one

- Fix and add some unit tests


lpegoraro (approved):
LGTM

PR #2144, opened 2023-01-11
[ORB-224] partial updates for datasets

mclcavalcante:
Support updates on datasets API to name and sinks ids without requiring the other fields to be provided in the JSON body:

Both cases follow the rules:

- If it is informed, the new one must overwrite the old one
- If it is omitted, the old one must be maintained.
- If it is informed with empty or invalid values, an error must be raised informing that the attribute doesn't match the pattern required.


lpegoraro (approved):
LGTM

PR #2149, opened 2023-01-17
[ORB-170] add 'noTapMatch' state for policies on agent

mclcavalcante:
Add a new policy state when a policy isn't applied because tap selector doesn't match any available taps
image


lpegoraro (approved):
LGTM

PR #2150, opened 2023-01-18
feat(migrate): add migrate service to default and support openTelemetry in all sinks

lpegoraro:
2 Agents, one OTEL=true and the other off.

Sinks Results:

## Agent's flag

image


weyrick (changes requested):
what about changing the docker provision commands in the UI too?

weyrick (approved):

PR #2162, opened 2023-01-27
[FIX] DeployCollector function

etaques:
This PR fixes DeployCollector function that was being called however it was unimplemented


manrodrigues (approved):
lgtm

PR #2163, opened 2023-01-28
[FIX] adding ownerID on decode GRPC sinks response

etaques:
This PR fixes the null parameter OwnerID that is returning on maestro when client.go is used to retrieveSinks from sinks micro service


manrodrigues (approved):
lgtm

PR #2164, opened 2023-01-29
[FIX] on logic for active and error sinks status

etaques:
This PR changes:
- Added logic to change LastRemoteWrite parameter for active sinks when receive a metric on sinker
- Symplify AnalyzeLogs function to check only errors in otel collector logs
- Removed LastRemoteWrite change on maestro monitorSinks function
- Just change sink status on maestro if it is active
- Fix deployment status on maestro micro service, should be internally controlled and not checking k8s deployment state, cuz can generate wrong status during update sink operation
- Remove otel idle state control from otel-sinker microservice, and added it on maestro with different manner

### DIFF
--- a/cmd/maestro/main.go
+++ b/cmd/maestro/main.go
@@ -11,13 +11,6 @@ package main
 import (
 	"context"
 	"fmt"
-	sinksgrpc "github.com/ns1labs/orb/sinks/api/grpc"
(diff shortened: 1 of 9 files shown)


lpegoraro (approved):
LGTM

lpegoraro (approved):

lpegoraro (approved):
LGTM

PR #2165, opened 2023-01-31
[FIX] refactoring maestro event sourcing for sink status

etaques:
This PR refactor maestro event sourcing creating its own redis stream and making sinks and sinker send events to it separately, instead maestro use sinks and sinker streams


lpegoraro (approved):
LGTM

PR #2171, opened 2023-02-01
[FIX] maestro event sourcing to update otel collector

etaques:
This fixes maestro update otel collector when sink has been updated.


manrodrigues (approved):
Lgtm

PR #2174, opened 2023-02-02
feat(migrate): fix migration applying all migrations, only last one is necessary

lpegoraro:
…p/down.


etaques (approved):
LGTM

PR #2176, opened 2023-02-03
fix(sinks): fix merging of the sinks tags and config.

lpegoraro:


etaques (approved):
LGTM

PR #2185, opened 2023-02-07
changing scenario descriptions for dry-run execution

manrodrigues:
Using dry-run execution to keep scenarios updated


etaques (approved):
LGTM

PR #2187, opened 2023-02-07
inserting tag on test scenarios with error status on sink

manrodrigues:
Those tests must pass locally using OTEL config on test_config. 
behavex -t @sink_status_error --parallel-processes=20 --parallel-scheme=scenario


lpegoraro (approved):
LGTM

PR #2188, opened 2023-02-07
fix(maestro): fix maestro reading events.

lpegoraro:


manrodrigues (approved):
Lgtm

PR #2192, opened 2023-02-09
fix(maestro): fix iteration to be more performatic and less error prone.

lpegoraro:


manrodrigues (approved):
lgtm

PR #2194, opened 2023-02-09
Orb 694 tests otel migration

manrodrigues:
- test agent legacy scraping through sink otel
- test agent legacy scraping through sink legacy updated to otel
- test agent legacy updated to otel scraping through sink legacy updated to otel
- test if all sinks are with otel enabled
- count sinks migrated
- test sink idle status

PS: it only will be valuable when we have status are reliable


lpegoraro (approved):
LGTM

PR #2198, opened 2023-02-10
fix(maestro): change from pod, to deployment and added logs for match…

lpegoraro:
…ing items


etaques (approved):
LGTM

PR #2204, opened 2023-02-13
feat(minikube): add minikube support for makefile

lpegoraro:
After a few issues with kind, I tried this yesterday at my pc, and worked better than kind.
I even tried using the new "skaffold" but didn't manage to work it out.
This didn't had the same issues that kind had, but faced a different issue about permissions that I believe is soon to be resolved.


etaques (approved):
LGTM

PR #2206, opened 2023-02-14
fix(redis): fixing how maestro and sinker handles errors in reading and handling event errors

lpegoraro:


etaques (approved):
LGTM

PR #2213, opened 2023-02-15
[ORB-673] Set default path for orb-agent.db

etaques:
This PR set the default path for orb-agent.db to /opt/orb/
image


lpegoraro (approved):
LGTM

PR #2222, opened 2023-02-16
 [FIX] set last activity on redis if sink already is active

etaques:
This PR set last activity on redis if sink already is active


lpegoraro (approved):
LGTM

PR #2223, opened 2023-02-16
Add otlp receiver v0.62 to orb agent

leoparente:


lpegoraro (approved):
LGTM

cmd/agent/main.go
```diff
@@ -146,6 +147,7 @@ func mergeOrError(path string) {
 	v.SetDefault("orb.db.file", "./orb-agent.db")
 	v.SetDefault("orb.tls.verify", true)
 	v.SetDefault("orb.otel.enable", false)
+	v.SetDefault("orb.otel.type", "prometheus")
```
weyrick, line 149: "type" is too generic, please make this "receiver_type" or similar. we do plan to deprecate this essentially immediately, right?

agent/backend/pktvisor/pktvisor.go
```diff
@@ -73,6 +78,7 @@ type pktvisorBackend struct {
 
 	// OpenTelemetry management
 	scrapeOtel bool
+	otelType   string
```
weyrick, line 81: otelReceiverType

agent/backend/pktvisor/scrape.go
```diff
@@ -156,9 +157,78 @@ func (p *pktvisorBackend) scrapeDefault() error {
 	return nil
 }
 
+func (p *pktvisorBackend) scrapeOtlp() {
+	policyID := "__all"
+	exeCtx, execCancelF := context.WithCancel(p.ctx)
+	exeCtx = context.WithValue(exeCtx, "policy_name", policyID)
```
weyrick, line 163: shouldn't this be a name, not ID? also elsewhere it looks like the policy name is just "policy", not "policy_name" so let's make sure we are consistent

agent/backend/pktvisor/scrape.go
```diff
@@ -156,9 +157,78 @@ func (p *pktvisorBackend) scrapeDefault() error {
 	return nil
 }
 
+func (p *pktvisorBackend) scrapeOtlp() {
```
weyrick, line 160: this opens up a port as a server, right? let's not call it "scrape" which implies that this will connect outwards, which it won't. maybe 'receiveOtlp'

agent/backend/pktvisor/scrape.go
```diff
@@ -156,9 +157,78 @@ func (p *pktvisorBackend) scrapeDefault() error {
...
+		count := 0
+		for {
+			if p.mqttClient != nil {
+				p.exporter[policyID], err = p.createOtlpMqttExporter(exeCtx, execCancelF)
+				if err != nil {
+					p.logger.Error("failed to create a exporter", zap.Error(err))
+					return
+				}
+				pFactory := otlpreceiver.NewFactory()
+				cfg := pFactory.CreateDefaultConfig()
```
weyrick, line 178: I think at least eventually we may need to be able to configure e.g. the port and path, perhaps not in this PR

weyrick (changes requested):

cmd/agent/main.go
leoparente: done. and you are right

agent/backend/pktvisor/scrape.go
leoparente: i change this. but kept policy_name to keep compatibility with current otel implementation. once we get this fully tested we can remove current otel code and rename it

agent/backend/pktvisor/scrape.go
leoparente: yes, you are right. done

agent/backend/pktvisor/scrape.go
leoparente: Agreed. I've created a issue for that: https://linear.app/netboxlabs/issue/ORB-709

weyrick (approved):

PR #2230, opened 2023-02-17
(feat/fix): Enable Partial Sink Update

gpazuch:
Add new component/screen for viewing and editing Sinks.

Allows user to edit sink's details and tags without changing or reentering sink backend configuration and password


gpazuch:
@manrodrigues couple of issues I'm fixing before merge:

- shows inexistent tag when no tags are present in the sink
- needs styling improvements:
  * fix each section dimensions
  * make it responsive
- should allow for backend change (? is/should it be possible to change backend ?)

manrodrigues:
> @manrodrigues couple of issues I'm fixing before merge:
> 
> * shows inexistent tag when no tags are present in the sink
> * needs styling improvements:
>   
>   * fix each section dimensions
>   * make it responsive
> * should allow for backend change (? is/should it be possible to change backend ?)

We'll have the otlp exporter that Leo is working on.. but I don't know how much responsible it should be or even if we should be able to change from one to another. We need to check it with Leo and Richard, probably.

gpazuch:
> @manrodrigues couple of issues I'm fixing before merge:
> 
> * shows inexistent tag when no tags are present in the sink

hey, I just noticed that for sinks created through the batch creation, there are tags being set as empty strings for key value pairs, like the one below

`{"id":"1da75ad8-d87e-405a-b1c3-767a8535cba2","name":"test_sink_label_name_AgHqDYgKLl","description":"","tags":{"":""},"state":"unknown","backend":"prometheus","config":{"migrated":"m3","opentelemetry":"enabled","password":"","remote_host":"https://prometheus-prod-10-prod-us-central-0.grafana.net/api/prom/push","username":"443932"},"ts_created":"2023-02-13T20:48:07.198133Z"}
`

image

PR #2243, opened 2023-02-21
fix(agent): fix renaming policy erroring and keep scraping

lpegoraro:


agent/backend/pktvisor/policy.go
```diff
@@ -61,7 +61,14 @@ func (p *pktvisorBackend) ApplyPolicy(data policies.PolicyData, updatePolicy boo
 func (p *pktvisorBackend) RemovePolicy(data policies.PolicyData) error {
 	p.logger.Debug("pktvisor policy remove", zap.String("policy_id", data.ID))
 	var resp interface{}
-	err := p.request(fmt.Sprintf("policies/%s", data.Name), &resp, http.MethodDelete, http.NoBody, "application/json", RemovePolicyTimeout)
+	var name string
+	// Since we use Name for removing policies not IDs, if there is a change, we need to remove the previous name of the policy
```
manrodrigues, line 65: @lpegoraro my thoughts here are that we should rename the policy on pktvisor - or remove and then add it again using the new name - all the time that a policy name is updated. And then we should use the newest name to remove it.
Considering this current approach you're proposing/using, my concern is this scenario:

1 - a policy named "policy-one" is applied to some group
2- this policy name is updated to "policy-updated"
3 - a new policy is created and receive the name "policy-one" (orb should allow it and also pktvisor, since the first policy is no longer named "policy-one", so there is no name conflict here.. but seems that it will not work properly)
4- And finally, which policy will be removed?

So, I guess the correct fix for it is to fix the update name and then use the last name to remove it (i'm almost sure that this is how it was used to work before)

what do you think?

manrodrigues (approved):
lgtm

PR #2260, opened 2023-03-13
feat(sinks): add support for YAML, following policies standard.

lpegoraro:
ORB-700


kind/values.yaml
```diff
@@ -1,6 +1,6 @@
 orb:
   defaults:
-    logLevel: "info"
+    logLevel: "debug"
```
manrodrigues, line 3: was that intentional?

sinks/api/http/endpoint_test.go
```diff
@@ -996,21 +1005,21 @@ func TestValidateSink(t *testing.T) {
...
-	}
-
-	for desc, tc := range cases {
-		t.Run(desc, func(t *testing.T) {
-			metadata := omitSecretInformation(tc.inputMetadata)
-			assert.Equal(t, tc.expectedMetadata, metadata)
-		})
-	}
-}
+//func TestOmitPasswords(t *testing.T) {
```
manrodrigues, line 1008: for what reason we are skipping this test?

kind/values.yaml
lpegoraro: it is, kind for local development, can be debug, we will store only local logs and it is purged after the deletion of the cluster

sinks/api/http/endpoint_test.go
lpegoraro: This is no longer applicable in this package, since now it has the dependency of the backend, I am looking for someplace else which is more suitable, to not add cyclic dependencies.

manrodrigues (approved):
lgtm

sinks/api/http/openapi.yaml
```diff
@@ -350,6 +364,39 @@ components:
...
+          type: string
+          example: prometheus
+          description: The sink backend to use. Must match a backend from /features/sinks. Cannot change once created.
+        format:
+          type: string
+          enum:
+            - yaml
+            - json
+          description: Format that the config will be passed on
+        config:
```
weyrick, line 397: this is missing config_data isn't it?

sinks/postgres/sinks.go
```diff
@@ -173,19 +173,21 @@ func (s sinksRepository) RetrieveAllByOwnerID(ctx context.Context, owner string,
...
 		return sinks.Page{}, errors.Wrap(errors.ErrSelectEntity, err)
 	}
 	defer rows.Close()
 
 	var items []sinks.Sink
 	for rows.Next() {
-		dbSink := dbSink{MFOwnerID: owner}
-		if err := rows.StructScan(&dbSink); err != nil {
+		d := dbSink{MFOwnerID: owner}
+		s.logger.Debug("iterating over rows", zap.Any("sink", d))
```
weyrick, line 185: we have to be careful here - I believe we have "debug" logging in orb.live at the moment. this should probably be removed unless it's really necessary.

sinks/sinks_service.go
```diff
@@ -71,13 +72,29 @@ func (svc sinkService) encryptMetadata(sink Sink) (Sink, error) {
...
+		newValue, err2 := svc.passwordService.EncodePassword(stringVal)
 		if err2 != nil {
 			err = err2
 			return key, value
 		}
 		return key, newValue
 	})
+	if sink.ConfigData != "" {
+		sinkBE := backend.GetBackend(sink.Backend)
+		sink.ConfigData, err = sinkBE.ConfigToFormat(sink.Format, sink.Config)
```
weyrick, line 91: probably should check that sinkBE is not nil

sinks/sinks_service.go
```diff
@@ -93,6 +110,14 @@ func (svc sinkService) decryptMetadata(sink Sink) (Sink, error) {
...
```
weyrick, line 115: here too

sinks/api/http/openapi.yaml
lpegoraro: good catch, I will change in my next PR!

sinks/sinks_service.go
lpegoraro: actually, at this point, the backend would have failed earlier.
when it is called by the `CreateSink`, it has a validation in the sink_service.go:35~38 iirc.
if it is called from `UpdateSink` the backend is already estabilished.

sinks/sinks_service.go
lpegoraro: > actually, at this point, the backend would have failed earlier. when it is called by the `CreateSink`, it has a validation in the sink_service.go:35~38 iirc. if it is called from `UpdateSink` the backend is already estabilished.


same reply

sinks/postgres/sinks.go
lpegoraro: yup, I will start adding TODOs in every debug log I add. I added those for local development, also, I don't think we need debug log in orb.live. is any debug log, really necessary in orb.live so we could check if we could go for info in those specific logs?

weyrick: i agree it would be better to make sure we are not on debug log level on orb.live. let's sync with @etaques

sinks/sinks_service.go
weyrick: understood, but I find the code will be more reliable if you add a simple check even here. code has a tendency to be e.g. copy/pasted to other places for example, where that may not be true.

sinks/postgres/sinks.go
lpegoraro: Commit with changes
https://github.com/orb-community/orb/pull/2268/commits/27add81715582d191df00103ad74607cef7d68e0

sinks/api/http/openapi.yaml
lpegoraro: Commit with changes
https://github.com/orb-community/orb/commit/27add81715582d191df00103ad74607cef7d68e0

sinks/sinks_service.go
lpegoraro: commit with fixes
https://github.com/orb-community/orb/pull/2268/commits/a0b7711187ad405b112fd2de0bbc4c4b38d85767

sinks/sinks_service.go
lpegoraro: commit with fixes
https://github.com/orb-community/orb/pull/2268/commits/a0b7711187ad405b112fd2de0bbc4c4b38d85767

PR #2261, opened 2023-03-13
Update agent.example.yaml

manrodrigues:
this came after my session with Xin.. the default policy failed to be applied because of the tap name.. so i think its worth keeping the pattern


lpegoraro (approved):
LGTM

PR #2274, opened 2023-03-20
fix(build:ui): read GTAG from environment variable

gpazuch:
Changed https://github.com/orb-community/orb/blob/develop/ui/scripts/setenv.ts to read GTAGID value from environment and write it to https://github.com/orb-community/orb/blob/develop/ui/src/environments/environment.env.ts upon build.

[] requires GTAGID to be defined on build environment


etaques (approved):
LGTM

PR #2277, opened 2023-03-22
fix(sinker): add debounce update to only change if the message of the…

lpegoraro:
… sink error is different from the previous error message.


etaques (approved):
LGTM

PR #2282, opened 2023-03-22
Revert "fix(sinker): add debounce update to only change if the messag…

lpegoraro:
…e of the sink error is different from the previous error message. (#2277)"

This reverts commit c7db52ad3d129135344b82e23cb9f4d47a7f0578.


etaques (approved):
lgtm

PR #2300, opened 2023-04-10
Clear otlp empty metrics if any

leoparente:


agent/otel/otlpmqttexporter/otlp.go
```diff
@@ -344,6 +349,7 @@ func (e *exporter) pushAllMetrics(ctx context.Context, md pmetric.Metrics) error
 		for key, value := range agentData.AgentTags {
 			scope = e.injectScopeAttribute(scope, key, value)
 		}
+		e.logger.Info("Scrapping policy via OTLP", zap.String("policyName", policyName))
```
leoparente, line 351: @manrodrigues I've added this log to help user and you to check policies that are being scrapped.. If you thing we should add a more meaningful or better message. Please, let me know

manrodrigues: this is good

PR #2305, opened 2023-04-12
feat(otel): add support for OTLP HTTP Exporter

lpegoraro:
Able to create Sinks with OTLP as Exporter
image

Maestro is able to spin up OtelCollector with OTLPExporter
image


sinks/sinks_service.go
```diff
@@ -62,71 +65,191 @@ func (svc sinkService) CreateSink(ctx context.Context, token string, sink Sink)
...
+func validateAuthType(s *Sink) (authentication_type.AuthenticationType, error) {
+	var authMetadata types.Metadata
+	if len(s.ConfigData) != 0 {
+		var helper types.Metadata
+		if s.Format == "yaml" {
+			err := yaml.Unmarshal([]byte(s.ConfigData), &helper)
+			if err != nil {
+				return nil, err
+			}
+			authMetadata = helper.GetSubMetadata("authentication")
```
weyrick, line 84: I would have liked to see this "authentication" string as a constant, rather than typing the string everywhere which is prone to typos

sinks/sinks_service.go
```diff
@@ -62,71 +65,191 @@ func (svc sinkService) CreateSink(ctx context.Context, token string, sink Sink)
...
 		if err != nil {
 			svc.logger.Error("error on parsing encrypted config in data")
 			return sink, err
 		}
+		sink.ConfigData = decodeMetadata.(string)
 	}
 	return sink, err
 }
 
+func (svc sinkService) UpdateSinkInternal(ctx context.Context, sink Sink) (Sink, error) {
```
weyrick, line 174: I believe this is only called during migration - if so, let's document that in the function comments here (and/or in the interface def)

maestro/config/exporter_builder.go
```diff
@@ -0,0 +1,44 @@
...
+	endpointCfg := config.GetSubMetadata("exporter")["remote_host"].(string)
+	return Exporters{
+		PrometheusRemoteWrite: &PrometheusRemoteWriteExporterConfig{
+			Endpoint: endpointCfg,
+			Auth:     Auth{Authenticator: authenticationExtensionName},
+		},
+	}, "prometheusremotewrite"
+}
+
+type OTLPExporterBuilder struct {
```
weyrick, line 33: should this be OTLPHTTPExporter, in case we add gRPC later?

sinks/backend/otlpexporter/configuration.go
```diff
@@ -0,0 +1,110 @@
+package otlpexporter
```
weyrick, line 1: again, shouldn't we name this after the otel exporter itself, otlphttp ?

sinks/postgres/init.go
```diff
@@ -75,6 +75,21 @@ func migrateDB(db *sqlx.DB) error {
...
+			{
+				Id: "sinks_3",
+				Up: []string{
+					`CREATE TABLE IF NOT EXISTS current_version (
+						id			     UUID NOT NULL DEFAULT gen_random_uuid(),
+						version          TEXT NOT NULL,
+    					last_updated     TIMESTAMPTZ DEFAULT CURRENT_TIMESTAMP NOT NULL,
+						PRIMARY KEY (id, version)
+					)`,
+					`INSERT INTO current_version  (id, version, last_updated) VALUES (DEFAULT, '0.25.0', DEFAULT);`,
```
weyrick, line 87: this "0.25.0" will force a migration i guess?

weyrick (changes requested):

weyrick (approved):

PR #2313, opened 2023-04-21
Change command to set mtu

etaques:
Due the problem with integration tests running on pipeline have better behavior using **ip link set dev** instead **ifconfig** to set mto for dummy interface

image


lpegoraro (approved):
LGTM

PR #2329, opened 2023-04-26
removing sink status check of flow without metrics

manrodrigues:
i'm removing sink status check of scenarios with flow without metrics.. sanity tests are now running on pipeline and we have validations with metrics there


leoparente (approved):
LGTM

PR #215, opened 2021-08-31
Feature/orb read policies | Unit test concurrency problem

dscabral:
Revolves #214, resolves #203, resolves #206

PR #2338, opened 2023-04-27
fix(feat): #215 add the policy status in the matching agents model

joaoguilherme2003:


ui/src/app/pages/fleet/agents/match/agent.match.component.ts
```diff
@@ -116,7 +131,16 @@ export class AgentMatchComponent implements OnInit, AfterViewInit {
     const tagsList = Object.keys(tags).map(key => ({ [key]: tags[key] }));
     this.agentsService.getAllAgents(tagsList).subscribe(
       resp => {
-        this.agents = resp;
+        if(!!this.policy){
+          this.agents = resp.map((agent)=>{
+            const {policy_state} = agent;
+            const policy_agg_info = !!policy_state && policy_state[this.policy.id].state || "Not Applied";
```
gpazuch, line 137: this is really neat!
'policy_agg_info' is overridden locally when we need to see specific policies in respect to the matching agents, but should display normally when policy context is absent.

gpazuch (approved):
:lgtm: :+1:

Issue #699, opened 2021-12-29
Run behave tests in GitHub CI


weyrick:
Daily behave test runs

Ensure appropriate docker container tags are tested, based on the branch.

manrodrigues:
Behave tests are integrated on our pipeline

https://qa.ops.orb.live/

PR #2349, opened 2023-05-02
feat(orb-ui): #699 Agent View, show how many policies are applied and how many are running

joaoguilherme2003:


gpazuch (approved):
:lgtm:

PR #2358, opened 2023-05-08
[ORB-174] (Fix) policy creation with json format 

brnvalenca:
This issue was occurring because inside the request.go file, in the validate() function, that is triggered by the addPolicyEndpoint, there was an if statement that converges to an error if a Policy is set and the Format of the policy was different from "". By this, if you set Policy (this means that is a json format) and specify the format as json (or any other) it would return an error.

The solution was changing the if statement to only accept req.Format as "json". And to trigger the error if json is not the format specified in the request and if the PolicyData is set (then it would be yaml not json).


etaques (approved):
LGTM

lpegoraro (approved):
LGTM

PR #2368, opened 2023-05-11
fix(orb-ui): edit sink through policy view

joaoguilherme2003:
Also added some styles that were missing in the View Sink page


gpazuch (approved):
:+1:

PR #2369, opened 2023-05-11
GET /sinks fail with 502 Bad Gateway code when sink is yaml format

brnvalenca:
It was returning 502 error code because it was trying to parse a string pointer to a string in the response. So i change some stuff in the ConfigToFormat function to pass directly the address of the string to make de response object. But if i do only that, then it would crash the create sink work flow. So i set up a flag to say what kind request is beign sent (post, put, get), and based on that i choose what type of conversion inside the ConfigToFormat it must be done.


kind/values.yaml
```diff
@@ -14,7 +14,7 @@ orb:
       name: "orb-sinks"
       pullPolicy: "IfNotPresent"
       repository: "orbcommunity"
-      tag: "develop"
+      tag: "0.25.0-044b0c82"
```
manrodrigues, line 17: why is this hard coded?

sinks/api/http/endpoint.go
```diff
@@ -50,14 +52,19 @@ func addEndpoint(svc sinks.SinkService) endpoint.Endpoint {
 			return nil, err
 		}
 
+		// Eu posso tentar setar uma flag para sinalizar qual o tipo do método que está
```
manrodrigues, line 55: Acho que a gente pode apagar isso aqui, ou mudar pro ingles, se for importante para voce manter o comentario :)

sinks/api/http/endpoint.go
```diff
@@ -125,7 +132,7 @@ func updateSinkEndpoint(svc sinks.SinkService) endpoint.Endpoint {
 			return nil, err
 		}
 		var config types.Metadata
-		if req.Format != "" {
+		if req.Format == "yaml" {
```
manrodrigues, line 135: is this modification necessary again? or your branch is not updated with develop?

manrodrigues (changes requested):

sinks/api/http/endpoint.go
brnvalenca: Boa amanda, isso é garbage. vou remover!

kind/values.yaml
brnvalenca: its wrong, i need to return to develop. forgot again... sorry. doing right now!

PR #2371, opened 2023-05-12
Reverting orb 727

manrodrigues:
This PR introduce this error https://linear.app/netboxlabs/issue/ORB-771/get-sinks-fail-with-502-bad-gateway-code-when-sink-is-yaml-format and do not solve the problem that was supposed to. For that reason, i'm reverting it.


brnvalenca (approved):
LGTM

PR #2392, opened 2023-05-31
feat(sinks): fix particular error which remote host is invalid.

lpegoraro:
This one is for you @manrodrigues


manrodrigues (approved):
lgtm

PR #2417, opened 2023-06-19
improvement(orb-ui): version filter agent list page

joaoguilherme2003:


gpazuch (approved):
:+1: some of the checks failed, so I set it to re-run all jobs - everything looks ok

PR #2455, opened 2023-06-25
feat: [ENG-745] - Added X-Scope-OrgID header

etaques:
This PR is permitting add X-Scope-OrgID on request header in otlphttpexporter and prometheusremotewrite through configuration below:
Prometheus remote writer:
image

Otlp http:
image

Here dashboard from otelcollector metrics pushing to mimir instance, which doesnt accept requests without **X-Scope-OrgID**
image

Here is mimir log showing header "tenant-2"
image


manrodrigues (approved):
Lgtm

maestro/config/exporter_builder.go
```diff
@@ -21,24 +25,73 @@ type PrometheusExporterConfig struct {
...
+	}
+
+	// Check to add X-Scope-OrgID header
+	headersMetadata := config.GetSubMetadata("headers")
+	if headersMetadata != nil {
+		if headerStr, ok := headersMetadata["X-Scope-OrgID"].(string); ok && headerStr != "" {
+			exporters.PrometheusRemoteWrite.Headers = make(map[string]string)
+			exporters.PrometheusRemoteWrite.Headers["X-Scope-OrgID"] = headerStr
+		}
+	}
```
lpegoraro, line 57: It will be better to not have the restriction to only this header, instead we can block the header we don't want.
And also, this is not the place to block, it should be on the sinks microservice.

sinks/headers_type/headers.go
lpegoraro, line 1: I don't think we are coming back to this format of form shipping, let's make it simple to only use the json or yaml format to configure sinks.

sinks/headers_type/scope/scope.go
lpegoraro, line 1: Again let's not use this method, the custom headers are not something separate from the authentication or the exporter. It is actually a exporter configuration we are adding, so everything should be inside the Exporter.

weyrick (commented):
Why do we need special support for "X-Scope-OrgID"? Why wouldn't this be generic support for any header?

weyrick (changes requested):
explain strategy

lpegoraro (changes requested):
I get it that it's working but the implementation is a bit off, we actually want a block list instead of a allow only list.
I will merge this with the branch I started this work and we can move along with this week.

maestro/config/exporter_builder.go
etaques: I also did restriction on sinks configuration thinking that will avoid peoples to use wrong headers in exporters request. The unique header that is used for prometheus like requests to control tenants. But I agree that have situations where we need other headers.

PR #2527, opened 2023-07-25
feat(sinker): add redundancy to create redis entry when notify active sink.

lpegoraro:


weyrick:
please add a description and note what problem this addresses

etaques (approved):
LGTM

PR #2662, opened 2023-09-27
feat(maestro): add observability and logs to new flow on maestro.

lpegoraro:
adding a few logs to handle messages, and add metrics to maestro to track how much messages it is receiveing and processing.


etaques (approved):
lgtm

etaques (approved):

PR #2676, opened 2023-10-02
fix(orb-ui): User able to create and edit yaml policies using json …

joaoguilherme2003:
also Include upload file policy view


gpazuch (approved):
:+1:

PR #2690, opened 2023-10-03
feat(maestro): fix get deployment and add logs to sink_activity.

lpegoraro:


maestro/deployment/service.go
```diff
@@ -169,15 +169,16 @@ func (d *deploymentService) UpdateDeployment(ctx context.Context, deployment *De
 	return nil
 }
 
-func (d *deploymentService) NotifyCollector(ctx context.Context, ownerID string, sinkId string, operation string, status string, errorMessage string) (string, error) {
-	got, deployName, err := d.GetDeployment(ctx, ownerID, sinkId)
```
etaques, line 173: here should be "manifest" instead "_"

etaques (changes requested):

etaques (approved):
LGTM

PR #2708, opened 2023-10-03
refactor: update readme and fix linting errors

gpazuch:
* update README
* fix .ts linting
* fix .scss linting
* fix missing imports on auth.component.scss
* add two new targets to package.json -> `lint:fix` and `lint:styles:fix`


joaoguilherme2003 (approved):
:100:

PR #2714, opened 2023-10-04
fix (maestro): reset error message when sink update

etaques:
we are not updating deployment table properly the record still in error state, thus this PR fix it


lpegoraro (changes requested):
wait a sec, I think I found the main issue.

PR #2715, opened 2023-10-04
fix(maestro): fix deployment skipping on last status was error

lpegoraro:


maestro/service/deploy_service.go
```diff
@@ -106,23 +106,14 @@ func (d *eventService) HandleSinkActivity(ctx context.Context, event maestroredi
 		return errors.New("trying to deploy sink that is not active")
 	}
 	d.logger.Debug("handling sink activity event", zap.String("sink-id", event.SinkID))
-	deploymentEntry, _, err := d.deploymentService.GetDeployment(ctx, event.OwnerID, event.SinkID)
```
etaques, line 109: this is correct, activity should ignore the errored sinks/ collectors

etaques (changes requested):

maestro/service/deploy_service.go
lpegoraro: reverted

etaques (approved):
lgtm

PR #2720, opened 2023-10-04
fix (maestro): activity and error event synchronization

etaques:
This PR changes:
- fixed updatedeployment function, was always getting old configuration
- fixed killotelcollector was only deleting deployment, not configmap and service
- fixed sinker activity stream subscribe and event consuption
- fixed sinker idle stream subscribe and event consuption
- fixed delete otel collector when sinks be idle


manrodrigues (approved):
Lgtm

PR #2739, opened 2023-10-09
feat(maestro): add redundancy to create deployment when receives acti…

lpegoraro:
…vity without deployment in the database.


etaques (approved):
lgtm

PR #2801, opened 2023-11-20
fix: python-tests - mkdocs-material 9.4.7 depends on mkdocs 1.5.3

mfiedorowicz:
Fix python-tests, `mkdocs-material` 9.4.7 depends on `mkdocs` 1.5.3


lpegoraro (approved):
LGTM

PR #2836, opened 2023-11-28
eng-1635-temp-mute-remotely-restart-scenario

manrodrigues:
we will back it to smoke tests as soon as it is fixed


mfiedorowicz (approved):
LGTM

PR #2837, opened 2023-11-28
eng-1632-temp-mute-2-agents-same-port

manrodrigues:
we will back it to smoke tests as soon as it is fixed


mfiedorowicz (approved):
Thank you

PR #2839, opened 2023-11-30
feat(agent): remove option to set otel to false in orb agent

lpegoraro:


mfiedorowicz (commented):
Do we need to keep `Opentelemetry.Enable` in https://github.com/orb-community/orb/blob/develop/agent/config/types.go#L39?

mfiedorowicz (approved):
LGTM

PR #2870, opened 2023-12-12
fix(agent): policy update and remove

mfiedorowicz:
Fixing following bugs:
- policy updates been not applied properly
- policy removals not fully handled (throwing errors around non-existing policies and temporary policy files not removed either)
- incorrect policy file paths


lpegoraro (approved):
LGTM

PR #2876, opened 2023-12-13
fix(agent): otel runner - break an infinite loop

mfiedorowicz:
On context cancelation, we must `return` regardless of the error, otherwise infinite loop causes high CPU usage.


manrodrigues (approved):
Lgtm

PR #2878, opened 2023-12-15
fix(otel): update otel collector in sinks and agent.

lpegoraro:
fixing partial updates and update agent.

### DIFF
--- a/agent/logging.go
+++ b/agent/logging.go
@@ -34,10 +34,10 @@ func (a *agentLoggerWarn) Printf(format string, v ...interface{}) {
 	a.a.logger.Warn("WARN mqtt log", zap.Any("payload", v))
 }
 func (a *agentLoggerDebug) Println(v ...interface{}) {
-	a.a.logger.Debug("DEBUG mqtt log", zap.Any("payload", v))
+	//a.a.logger.Debug("DEBUG mqtt log", zap.Any("payload", v))
 }
 func (a *agentLoggerDebug) Printf(format string, v ...interface{}) {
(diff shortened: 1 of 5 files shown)


agent/logging.go
```diff
@@ -34,10 +34,10 @@ func (a *agentLoggerWarn) Printf(format string, v ...interface{}) {
 	a.a.logger.Warn("WARN mqtt log", zap.Any("payload", v))
 }
 func (a *agentLoggerDebug) Println(v ...interface{}) {
-	a.a.logger.Debug("DEBUG mqtt log", zap.Any("payload", v))
+	//a.a.logger.Debug("DEBUG mqtt log", zap.Any("payload", v))
```
mfiedorowicz, line 37: Can we re-add these, please?

agent/logging.go
```diff
@@ -34,10 +34,10 @@ func (a *agentLoggerWarn) Printf(format string, v ...interface{}) {
...
 }
 func (a *agentLoggerDebug) Printf(format string, v ...interface{}) {
-	a.a.logger.Debug("DEBUG mqtt log", zap.Any("payload", v))
+	//a.a.logger.Debug("DEBUG mqtt log", zap.Any("payload", v))
```
mfiedorowicz, line 40: Same here

mfiedorowicz (commented):
Don't we need to also update:
- https://github.com/orb-community/orb/blob/develop/fleet/backend/otel/types.go#L3
- version in go.mod
?

PR #2879, opened 2023-12-16
fix: otel backend integration tests

mfiedorowicz:
Fixing integration tests for otel backend (also muted 3 tests as having same issues as for pktvisor from the past) + one typo.


manrodrigues (approved):
Lgtm

PR #2881, opened 2023-12-18
feat(POC): otelcol httpcheck receiver - use tags as metrics labels

mfiedorowicz:
To allow using `tags` in the httpcheck receiver targets as metrics labels and be able to query metrics by common tags.

Example target definition:
```json
{
  "endpoint": "https://orb.community",
  "method": "GET",
  "tags": {
    "tag_a": "valueA",
    "tab_b": "valueB"
}
```

Main issue with this is if the policy having httpcheck target with `tags` is being used with older orb agent not having policy sanitization, it will end up with validation error on otelcol and ending up policy not being applied.

### DIFF
--- a/agent/backend/otel/policy.go
+++ b/agent/backend/otel/policy.go
@@ -2,6 +2,7 @@ package otel
 
 import (
 	"context"
+	"encoding/json"
 	"errors"
 	"fmt"
 	"os"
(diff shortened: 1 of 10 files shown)


agent/backend/otel/policy.go
```diff
@@ -175,3 +184,37 @@ func (o *openTelemetryBackend) ValidatePolicy(otelConfig openTelemetryConfig) er
 
 	return nil
 }
+
+func SanitizePolicyData(policyData policies.PolicyData) (*policies.PolicyData, error) {
+	originalJSON, err := json.Marshal(policyData)
+	if err != nil {
+		return nil, err
+	}
+	policyDataClone := policies.PolicyData{}
```
lpegoraro, line 193: just out for standards, why not use the var here?

mfiedorowicz: Good point

PR #2894, opened 2023-12-20
fix(sinks): prevent circular config metadata encryption

mfiedorowicz:
We discovered an issue in the sinks service when user updates existing sink (i.e. name, description, tags) without config. In such scenario, encrypted password of the existing config metadata is being encrypted again. This leads to existing sink/otelcollectors not being able to authenticate with prometheus/otlphttp backends. 

For sinks already affected by this issue, the recommended solution is to update existing sinks with correct backend credentials.


manrodrigues (approved):
lgtm

lpegoraro (approved):
LGTM

PR #2919, opened 2024-01-02
fix(agent): remove scraper it was used on legacy code, cleaning up.

lpegoraro:
Failed test suite is passing now.

image

- [x] Re-running Smoke suite.
image

### DIFF
--- a/agent/backend/pktvisor/pktvisor.go
+++ b/agent/backend/pktvisor/pktvisor.go
@@ -1,6 +1,4 @@
-/* This So	if p.scraper != nil {
-		p.scraper.Stop()
-	}urce Code Form is subject to the terms of the Mozilla Public
+/* This Source Code Form is subject to the terms of the Mozilla Public
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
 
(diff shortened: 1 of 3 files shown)


agent/backend/pktvisor/pktvisor.go
```diff
@@ -1,4 +1,6 @@
-/* This Source Code Form is subject to the terms of the Mozilla Public
+/* This So	if p.scraper != nil {
+		p.scraper.Stop()
+	}urce Code Form is subject to the terms of the Mozilla Public
```
mfiedorowicz, line 3: Not sure if that's intended :)

agent/backend/pktvisor/pktvisor.go
```diff
@@ -17,7 +19,6 @@ import (
 
 	mqtt "github.com/eclipse/paho.mqtt.golang"
 	"github.com/go-cmd/cmd"
-	"github.com/go-co-op/gocron"
```
mfiedorowicz, line 20: Are we using it for anything else? If not, go dependency should be removed from go.mod

lpegoraro: was waiting the results of the report to send the clean up. here you go

mfiedorowicz (approved):

PR #2920, opened 2024-01-02
fix(agent): remove policy id as tap.

lpegoraro:


agent/backend/otel/policy.go
```diff
@@ -83,7 +83,6 @@ func (o *openTelemetryBackend) ApplyPolicy(newPolicyData policies.PolicyData, up
 				return err
 			}
 
-			o.otelReceiverTaps = append(o.otelReceiverTaps, newPolicyData.ID)
```
mfiedorowicz, line 86: Could we also remove that redundant empty line above removed append, please?

mfiedorowicz (approved):

PR #2921, opened 2024-01-08
fix(sinks): add default format as json.

lpegoraro:
image

### DIFF
--- a/sinks/postgres/sinks.go
+++ b/sinks/postgres/sinks.go
@@ -397,7 +397,11 @@ func toSink(dba dbSink) (sinks.Sink, error) {
 		configData = *dba.ConfigData
 	}
 	if dba.Format != nil {
-		format = *dba.Format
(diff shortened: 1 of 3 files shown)


sinks/postgres/sinks.go
```diff
@@ -397,7 +397,13 @@ func toSink(dba dbSink) (sinks.Sink, error) {
...
 	}
 	if dba.Format != nil {
-		format = *dba.Format
+		if len(*dba.Format) == 0 {
+			format = "json"
+		} else {
+			format = *dba.Format
+		}
+	} else {
+		format = "json"
```
mfiedorowicz, line 406: If we're defaulting format to `json`, this logic can be simplified i.e.

```go
format := "json"

if dba.Format != nil && *dba.Format != "" {
	format = *dba.Format
}
``

mfiedorowicz (approved):
LGTM 👍🏻

PR #2985, opened 2024-02-01
chore: bulk dependabot fixes

mfiedorowicz:


lpegoraro (approved):
LGTM, I usually do the cherry-pick to keep the original commit, but that is ok too.

PR #2986, opened 2024-02-01
fix(agent): warn instead of error the policy removal error when polic…

lpegoraro:
…y was not found.

### DIFF
--- a/agent/backend/pktvisor/policy.go
+++ b/agent/backend/pktvisor/policy.go
@@ -8,6 +8,7 @@ import (
 	"bytes"
 	"fmt"
 	"net/http"
+	"strings"
 
 	"github.com/orb-community/orb/agent/policies"
 	"go.uber.org/zap"
@@ -60,8 +61,11 @@ func (p *pktvisorBackend) RemovePolicy(data policies.PolicyData) error {
 	} else {
 		name = data.Name
 	}
-	err := p.request(fmt.Sprintf("policies/%s", name), &resp, http.MethodDelete, http.NoBody, "application/json", RemovePolicyTimeout)
-	if err != nil {
+	if err := p.request(fmt.Sprintf("policies/%s", name), &resp, http.MethodDelete, http.NoBody, "application/json", RemovePolicyTimeout); err != nil {
+		if strings.Contains(err.Error(), "404") {
+			p.logger.Warn("ignoring error from removing a policy which was not found", zap.String("policy_id", data.ID), zap.String("policy_name", name))
+			return nil
+		}
 		return err
 	}
 	return nil


agent/backend/pktvisor/policy.go
```diff
@@ -60,8 +61,11 @@ func (p *pktvisorBackend) RemovePolicy(data policies.PolicyData) error {
 	} else {
 		name = data.Name
 	}
-	err := p.request(fmt.Sprintf("policies/%s", name), &resp, http.MethodDelete, http.NoBody, "application/json", RemovePolicyTimeout)
-	if err != nil {
+	if err := p.request(fmt.Sprintf("policies/%s", name), &resp, http.MethodDelete, http.NoBody, "application/json", RemovePolicyTimeout); err != nil {
+		if strings.Contains(err.Error(), "404") {
+			p.logger.Warn("ignoring error from removing a policy which was not found", zap.String("policy_id", data.ID), zap.String("policy_name", name))
```
mfiedorowicz, line 66: s/ignoring error from removing a policy which was not found/policy not found

agent/otel/otlpmqttexporter/otlp.go
```diff
@@ -196,6 +196,13 @@ func (e *baseExporter) injectScopeLogsAttribute(logsScope plog.ScopeLogs, attrib
 	return logsScope
 }
 
+func (e *baseExporter) shutdown(_ context.Context) error {
+	if e.config.Client == nil || !(*e.config.Client).IsConnected() {
+		(*e.config.Client).Disconnect(0)
```
mfiedorowicz, line 201: Won't it panic if `e.config.Client` is `nil`?

mfiedorowicz: Also, why would we need to disconnect if it's not connected?

agent/comms.go
```diff
@@ -27,8 +28,12 @@ func (a *orbAgent) connect(ctx context.Context, config config.MQTTConfig) (mqtt.
 	})
 	opts.SetConnectionLostHandler(func(client mqtt.Client, err error) {
 		a.logger.Error("connection to mqtt lost", zap.Error(err))
+		// If it is a bug on the mqttclient, stop the agent
+		if strings.Contains(err.Error(), "BUG") {
+			a.Stop(ctx)
+		}
 		a.logger.Info("reconnecting....")
-		client.Connect()
+		a.requestReconnection(ctx, a.client, config)
```
mfiedorowicz, line 36: `a.Stop(ctx)` stops all configured backends, don't we need to start the agent again in this scenario to setup backend to the initial state?

lpegoraro: the a.Stop is a termination command, I can leave it explicitly with a return command there!

agent/otel/otlpmqttexporter/otlp.go
lpegoraro: By Disconnecting the handler of disconnection will kick in, and do the retry to connect. And yeah, the logic was inverted there.

agent/backend/pktvisor/policy.go
```diff
@@ -60,8 +61,11 @@ func (p *pktvisorBackend) RemovePolicy(data policies.PolicyData) error {
 	} else {
 		name = data.Name
 	}
-	err := p.request(fmt.Sprintf("policies/%s", name), &resp, http.MethodDelete, http.NoBody, "application/json", RemovePolicyTimeout)
-	if err != nil {
+	if err := p.request(fmt.Sprintf("policies/%s", name), &resp, http.MethodDelete, http.NoBody, "application/json", RemovePolicyTimeout); err != nil {
+		if strings.Contains(err.Error(), "404") {
+			p.logger.Warn("ignoring error from removing a policy not found", zap.String("policy_id", data.ID), zap.String("policy_name", name))
```
mfiedorowicz, line 66: Can we change warning message to `policy not found`?

PR #2993, opened 2024-02-07
feat(opentelemetry): use custom otelcol-orb collector for orb-agent and orb-maestro

lpegoraro:
Blocked by
- [x] Merge [PR](https://github.com/orb-community/otelcol-orb/pull/1)
- [ ] Images on DockerHub orbcommunity/otelcol-orb-agent and orbcommunity/otelcol-orb-maestro


maestro/config/config_builder.go
```diff
@@ -86,7 +86,7 @@ var k8sOtelCollector = `
             "containers": [
               {
                 "name": "otel-collector",
-                "image": "otel/opentelemetry-collector-contrib:0.91.0",
+                "image": "orb-community/otelcol-orb-maestro@latest",
```
mfiedorowicz, line 89: I'd recommend to avoid using `latest` but stick to specific version (once we have them)

maestro/config/config_builder.go
```diff
@@ -296,7 +296,7 @@ var JsonDeployment = `
...
```
mfiedorowicz, line 299: Same here regarding `latest`

.github/workflows/agent.yml
```diff
@@ -218,3 +218,83 @@ jobs:
       - name: Push agent container orbcommunity
         if: github.event_name != 'pull_request'
         run: docker push -a orbcommunity/orb-agent-full
+
+  package-agent-with-otelcol-orb:
```
mfiedorowicz, line 222: Is it a temporary step until we verify otelcol-orb works as expected and then swap orb-agent-full instead?

sinker/otel/bridgeservice/bridge.go
```diff
@@ -21,7 +21,7 @@ type BridgeService interface {
 	GetDataSetsFromAgentGroups(ctx context.Context, mfOwnerId string, agentGroupIds []string) (map[string]string, error)
 	NotifyActiveSink(ctx context.Context, mfOwnerId, sinkId, state, message string) error
 	GetSinkIdsFromPolicyID(ctx context.Context, mfOwnerId string, policyID string) (map[string]string, error)
-	IncreamentMessageCounter(publisher, subtopic, channel, protocol string)
+	IncrementMessageCounter(publisher, subtopic, channel, protocol string)
```
mfiedorowicz, line 24: Nice catch 🔍

.github/workflows/agent.yml
lpegoraro: I would have both for now and then swap after testing.

